PR #24, opened 2016-04-19
Change reboot task to just wait for host to come up

miabbott:
I kept running into trouble with the reboot task when testing with local
VMs.  Sometimes the VM wouldn't be identified as having come down, other
times the VM wouldn't be identified as having come up.

After some experimenting, I settled on just waiting for the host to come
up after a short delay.

### DIFF
--- a/common/ans_reboot.yaml
+++ b/common/ans_reboot.yaml
@@ -8,7 +8,10 @@
 #       To mitigate this, we prefix a sleep command before the shutdown so
 # ansible has time to move on. For more info on this issue, see:
 # https://github.com/ansible/ansible/issues/10616
-
+#
+# The Ansible docs now recommend this combination of tasks to handle reboots
+# https://support.ansible.com/hc/en-us/articles/201958037-Reboot-a-server-and-wait-for-it-to-come-back
+#
 - name: restart hosts
   shell: sleep 3 && shutdown -r now
   async: 1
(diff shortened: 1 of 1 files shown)


cgwalters:
Tricky...indeed there is a race with the old code because the host might come up faster than we detect the ssh socket is down.  But now it seems like there might be a new race where if we poll for the ssh socket faster than `sshd.service` shuts down, we might spuriously find it up.

I guess with `delay=30` we're avoiding any races, but that means we have a 30s speed hit.

I bet it would work to do `systemctl stop sshd && systemctl reboot` - ssh is set up so the current connection stays alive even though the main daemon stops listening.

miabbott:
I'm using the paradigm that the Ansible docs suggest - 

https://support.ansible.com/hc/en-us/articles/201958037-Reboot-a-server-and-wait-for-it-to-come-back

But I will experiment with `systemctl stop sshd && systemctl reboot` to see how well it works

cgwalters:
Yeah, probably simplest to just do what the docs suggest.  I think it'd mainly be useful to fix for making it faster to run the tests locally.

miabbott:
:arrow_up: 

Reduced the delay to 15s which seemed to work well using local VMs and OpenStack VMs.  And added some comments to the task about this being the recommended way of doing things.

jlebon:
Totally missed that convo (thought I was subscribed!). Yeah, that makes sense overall. Another approach here if we want as-fast-as-possible response time, would be a script that does e.g.:

```
locally: ( wait for port 22 down && wait for port 22 up ) &
on remote: systemctl reboot
```

By starting the wait_for_down event *before* rebooting, we eliminate the race condition.

PR #26, opened 2016-04-20
Add support for testing new tree via rebase

miabbott:
This change introduces the ability to perform the smoketest on a tree
that was delivered via `rpm-ostree rebase`.  In doing so, a number of
tags were introduced for some tasks in order to limit which delivery
mechanism can be used.

For example, when testing a new tree delivered via `rpm-ostree upgrade`,
the test should not run the `rpm-ostree rebase` command (and associated
tasks).  This is accomplished through the use of `--skip-tags
rpm_ostree_rebase` when running the playbook.  A similar exclusion
strategy is used when testing a new tree delivered via `rpm-ostree
rebase`.

This does increase the complexity when running the playbook, but also
increases flexibility.

It would probably be worthwhile to investigate how to simplify this test
by divorcing the choice of delivery (`upgrade` vs. `rebase`) from the
test itself.

### DIFF
--- a/tests/new-tree-smoketest/README.md
+++ b/tests/new-tree-smoketest/README.md
@@ -1,6 +1,7 @@
 This playbook performs a simple test of the upgrade and rollback mechanisms
-that are present in an Atomic Host.  This playbook should run on the Atomic
-Host variants of CentOS, Fedora, and Red Hat Enterprise Linux.
+that are present in an Atomic Host.  This playbook also supports testing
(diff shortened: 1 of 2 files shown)


miabbott:
I guess since I'm currently the only consumer of these tests, I can safely merge this.  It's been over a week after all.

cgwalters:
LGTM.  We need a review-assigner bot.

PR #27, opened 2016-05-10
Add a Vagrantfile to run the new-image smoketest

cgwalters:
We should make it convenient for people to test this
locally, similar to what openshift-ansible has.

This is just a stub file to run the new-image-smoketest - it currently
fails, looks like because we're missing a template file.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,18 @@
+# vi: set ft=ruby :
+
+# In the future, this might do something more interesting
+# like contain code to sync git from the local dir into
+# the system and do builds/installs there.  But for
+# now at least this exists as a starting point, and
+# once `ostree admin unlock` exists in the next version,
+# things will be a bit simpler.
+
+Vagrant.configure(2) do |config|
+    config.vm.box = "centos/atomic-host"
+    config.vm.hostname = "centosah-dev"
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -0,0 +1,18 @@
...
+# once `ostree admin unlock` exists in the next version,
+# things will be a bit simpler.
+
+Vagrant.configure(2) do |config|
+    config.vm.box = "centos/atomic-host"
+    config.vm.hostname = "centosah-dev"
+    config.vm.synced_folder ".", "/srv/vagrant", disabled: true
+
+    config.vm.provision "ansible" do |ansible|
+      ansible.playbook = "tests/new-image-smoketest/main.yaml"
```
miabbott: You can add

`ansible.skip_tags = "jenkins"`

...here and that should skip any of the templating tasks and avoid any failures you may have seen.

miabbott:
I did a `vagrant up` and didn't see any template errors.  See my note about how the tests can skip those tasks (not critical to the test path itself).

Otherwise LGTM

miabbott:
I did a fixup to add the 'skip_tags' portion and merged from the command line.

First time doing this, so hopefully it was done correctly.

PR #29, opened 2016-05-27
tests: new unique-machine-id playbook

miabbott:
This playbook verifies that the contents of `/etc/machine-id` are
unique across multiple hosts.  It requires exactly two hosts that have
the same distribution and distribution version.

With this change, an attempt was made to start following the Openshift
Ansible best practices, as mentioned in issue #1.  This includes
checking that variables are defined before playbook/role/task
execution and a shift away from using `include:` statements.

The new approach is to use roles and symlink a top-level directory
into each individual test directory.  This has been found to work
quite well.


miabbott:
@mnguyen-gh have a look

mnguyen-gh:
Ansible recommends the following naming scheme for the roles.  I'm not sure if it makes sense to incorporate that somehow into the naming of the roles in our tests.  If not we can come up with our own naming schema to have consistency in our roles. 

i.e. compare_machine_id   ->  machine_id_comparison

[https://github.com/openshift/openshift-ansible/blob/master/docs/best_practices_guide.adoc#ansible](url)
> Ansible Roles SHOULD be named like technology_component[_subcomponent].
For consistency, role names SHOULD follow the above naming pattern. It is important to note that this is a recommendation for role naming, and follows the pattern used by upstream.
Many times the technology portion of the pattern will line up with a package name. It is advised that whenever possible, the package name should be used.
Examples:
The role to configure a master is called openshift_master
The role to configure OpenShift specific yum repositories is called openshift_repos

miabbott:
@mnguyen-gh Updated this PR with naming that is more in line with the OpenShift Ansible style

mnguyen-gh:
@miabbott Looks good!

miabbott:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;cucbioWW&#x2F;252-2-migrate-unique-machine-id-test-upstream"> (2) Migrate unique machine ID test upstream</a>

PR #36, opened 2016-07-15
add RPM db tests

miabbott:
This introduces a simple test of the RPM database to the sanity suite,
among other things.  Other changes:
  - the `Vagrantfile` now has Fedora 24 support
  - the playbook executed by the Vagrant box can be configured by an
    environment variable
  - the `rpm-ostree upgrade` role has been improved to check for an
    available update first

### DIFF
--- a/roles/rpmdb_verify/tasks/main.yml
+++ b/roles/rpmdb_verify/tasks/main.yml
@@ -8,4 +8,4 @@
 - name: Fail if the RPM database is broken
   fail:
     msg: "The RPM database is not usuable"
-  when: rpmdb.stdout <= 0 or rpmdb.rc != 0
+  when: "{{ rpmdb.stdout | int }} <= 0 or rpmdb.rc != 0"
--- a/roles/rpmdb_verify/tasks/main.yml
+++ b/roles/rpmdb_verify/tasks/main.yml
@@ -1,0 +1,11 @@
+---
+# vim: set ft=ansible:
+#
+- name: Verify that the RPM database is intact
+  shell: rpm -qa | wc -l
+  register: rpmdb
+
+- name: Fail if the RPM database is broken
+  fail:
+    msg: "The RPM database is not usuable"
+  when: rpmdb.stdout <= 0 or rpmdb.rc != 0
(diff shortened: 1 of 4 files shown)


roles/rpmdb_verify/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
...
+# vim: set ft=ansible:
+#
+- name: Verify that the RPM database is intact
+  shell: rpm -qa | wc -l
+  register: rpmdb
+
+- name: Fail if the RPM database is broken
+  fail:
+    msg: "The RPM database is not usuable"
+  when: rpmdb.stdout <= 0 or rpmdb.rc != 0
```
cgwalters: Hmmm...don't you have to convert `stdout` explicitly into an integer from a string or something?

miabbott: Probably a wise idea either way.

roles/rpmdb_verify/tasks/main.yml
```diff
@@ -8,4 +8,4 @@
 - name: Fail if the RPM database is broken
   fail:
     msg: "The RPM database is not usuable"
-  when: rpmdb.stdout <= 0 or rpmdb.rc != 0
+  when: "{{ rpmdb.stdout | int }} <= 0 or rpmdb.rc != 0"
```
cgwalters: I guess it seems to me we want something like `len(rpmdb.stdout)` ?

miabbott: The output should be a number; I'm using `rpm -qa | wc -l` as the command, so assuming it returns successfully, I believe the integer comparison should work.

cgwalters: Yep you're right, sorry!

miabbott:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;5Jkt0oas&#x2F;353-2-expand-a-h-t-sanity-suite-to-include-rpm-db-check"> (2) Expand a-h-t sanity suite to include RPM db check</a>

PR #38, opened 2016-07-20
roles: use ansible_hostname instead of ansible_fqdn

miabbott:
While using the `improved_sanity_test` playboook in our OpenStack
environment (when driven by Jenkins), we were seeing failures when trying to
verify changes made to `/etc/hosts`.  (We did not see the same
failures when running the playbook by hand.)

The root cause was the value of `ansible_fqdn` was changing after the
host was rebooted.  I don't fully understand why this was changing
after a reboot, but in order to work around this, we can just use
`ansible_hostname` in its stead.

### DIFF
--- a/roles/etc_verify_changes/tasks/main.yml
+++ b/roles/etc_verify_changes/tasks/main.yml
@@ -2,7 +2,7 @@
 # vim: set ft=ansible:
 #
 - name: Verify changes to /etc/hosts
-  command: grep "{{ ansible_default_ipv4.address }} {{ ansible_hostname}}" /etc/hosts
+  command: grep "{{ ansible_default_ipv4.address }} {{ ansible_hostname }}" /etc/hosts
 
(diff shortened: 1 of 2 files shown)


roles/etc_verify_changes/tasks/main.yml
```diff
@@ -2,7 +2,7 @@
 # vim: set ft=ansible:
 #
 - name: Verify changes to /etc/hosts
-  command: grep "{{ ansible_default_ipv4.address }} {{ ansible_fqdn }}" /etc/hosts
+  command: grep "{{ ansible_default_ipv4.address }} {{ ansible_hostname}}" /etc/hosts
```
mnguyen-gh: Add space after ansible_hostname

miabbott:
@mnguyen-gh :arrow_up: fixed

mnguyen-gh:
Ran it through OpenStack instance with Jenkins and it passed.

PR #40, opened 2016-08-09
Adding info.txt which has the description of the tests that are run

mike-nguyen:
This is a basic text file with summary of tests in playbook.  It will help users understand what the playbook does and will also be used in CI notifications.


miabbott:
:shipit:

PR #41, opened 2016-08-12
Add cockpit test to improve sanity test

mike-nguyen:


mike-nguyen:
@miabbott have a look when you get the chance

miabbott:
LGTM, but could you update the commit message to have a little more description of what it does.

roles/cockpit_run_verify/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+    cockpit_cname: "cockpit/ws"
+  when: ('CentOS' in ansible_distribution) or ansible_distribution == "Fedora"
+
+- name: Run cockpit container
+  command: atomic run {{ cockpit_cname }}
+
+- name: Verify port 9090 is open
+  wait_for:
+    port: 9090
+    timeout: 120
```
miabbott: Actually, this seems a bit long.  Is there a reason we need to wait so long?

mike-nguyen: It should start up right away.  Do you think 30 seconds would be good before it fails?

miabbott: Yeah, that should be plenty of time.

mike-nguyen:
Timeout updated to 30 seconds.

roles/cockpit_run_verify/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+    cockpit_cname: "registry.access.redhat.com/rhel7/cockpit-ws"
+  when: ansible_distribution == "RedHat"
+
+- name: Set cockpit container name for Fedora/CentOS
+  set_fact:
+    cockpit_cname: "cockpit/ws"
+  when: ('CentOS' in ansible_distribution) or ansible_distribution == "Fedora"
+
+- name: Run cockpit container
+  command: atomic run {{ cockpit_cname }}
```
cgwalters: I'd separate `docker pull {{ cockpit_cname }}` from `atomic run`.  And have a separate timeout (at least a minute for download, but just ~10 seconds for run?

mike-nguyen: @cgwalters The timeout only applies to the waiting for port 9090 to open.  In general, we have not explicitly set timeouts for commands to run in Ansible.  This is just a simple test to verify the cockpit container can be run.  

Is there a reason why you want to see a separate docker pull and atomic run?  There will be specific docker and atomic tests added to the improved sanity tests in the near future so we will have coverage for the individual commands.  @miabbott is working on the docker sanity tests right now.

cgwalters: Yep, misunderstood.  LGTM as is.

miabbott:
Time to :shipit:

miabbott:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;S2uqM9YQ&#x2F;380-2-add-cockpit-test-to-a-h-t-sanity-suite"> (2) Add cockpit test to a-h-t sanity suite</a>

PR #42, opened 2016-08-15
improved-sanity-test: Add tests for docker operations

miabbott:
This brings in some simple docker tests to the sanity suite.  There
are tests performed pre and post upgrade of the system.

Before the upgrade happens, a base image is pulled to be used as part
of the build process of a layered image.  The layered image is built,
run, and then the resulting container is removed.

After the upgrade, the layered image that was built is ran again, the
resulting container removed, and then the layered image is removed.

As part of this change, I also pinned the playbook to Ansible 2.1 as
it seems pretty stable and has been working well thusfar.

### DIFF
--- a/roles/docker_build_httpd/files/centos_httpd_Dockerfile
+++ b/roles/docker_build_httpd/files/centos_httpd_Dockerfile
@@ -1,0 +1,16 @@
+FROM centos
+MAINTAINER Micah Abbott <micah@redhat.com>
+
+LABEL Version=1.1
+LABEL RUN="docker run -d --name NAME -p 80:80 IMAGE"
+
+ENV container docker
+
+RUN yum -y install httpd && \
+    yum clean all
+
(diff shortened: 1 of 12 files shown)


miabbott:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;MIuItFIM&#x2F;354-3-expand-a-h-t-sanity-suite-to-include-docker-tests"> (3) Expand a-h-t sanity suite to include Docker tests</a>

cgwalters:
LGTM :+1: 

Down the line though, I wonder if we should be using httpd from [CentOS Dockerfiles](https://github.com/CentOS/CentOS-Dockerfiles) but this is fine for now.

miabbott:
Opened issue #43 for the Dockerfile suggestion

mike-nguyen:
LGTM

PR #33, opened 2016-07-07
Improved sanity tests

miabbott:
The existing sanity test(s) have always felt lacking in terms of
functional tests that are executed.  This change attempts to improve
(or perhaps, replace) the existing sanity test(s).

This is the first cut at these types of tests.  The intent is to
expand the coverage of this set of tests to include additional tests
that are small and focused.  More complex functional tests can be
covered in separate playbooks which can be run after this.

Notable changes:
  - roles everywhere
  - hard commitment to Ansible version 2+
  - selectable Vagrant boxes for testing playbooks
  - verify that `/tmp` is functional after each reboot
  - verify that added users persist after upgrade
  - verify changes to `/etc` persist after upgrade
  - verify that directories/files in `/var` persist after
    upgrade/rollback
  - verify that users added after upgrade do not persist after
    rollback

PR #44, opened 2016-08-19
roles: remove ansible_host reboot case

miabbott:
During recent testing of incoming changes, some of us have noticed
inconsistency when using the `ansible_host` variable during the
'reboot' role.  Instead, the `inventory_hostname` variable has given
us more more reliability.

This somewhat flies in the face of the comment originally made in
PR #33, but live and learn.

### DIFF
--- a/roles/reboot/tasks/main.yml
+++ b/roles/reboot/tasks/main.yml
@@ -22,19 +22,6 @@
 # that they don't use sudo, which may require a password, and is not necessary
 # anyway.
 
-# During playbook testing, I found that sometimes the 'ansible_host' variable
-# is sometimes undefined, leading to errors waiting for the host to return.
-# We can opt to use the 'inventory_hostname' in this case and avoid those
-# (admittedly) rare error cases.
-- name: wait for hosts to come back up (ansible_host)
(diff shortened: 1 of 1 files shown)


roles/reboot/tasks/main.yml
```diff
@@ -22,24 +22,10 @@
...
-  local_action: wait_for host={{ ansible_host }}
-                port=22
-                state=started
-                delay=15
-                timeout=120
-  become: false
-  when: ansible_host is defined
-
 - name: wait for hosts to come back up (inventory_hostname)
   local_action: wait_for host={{ inventory_hostname }}
```
mike-nguyen: I have never seen inventory_hostname undefined, but if you want to keep ansible_host as a backup you can use:

host="{{ ansible_host | default(inventory_hostname) }}"

Other than this, LGTM.

miabbott: The problems we have observed are when `ansible_host` is defined, so I don't want to use it at all, if possible.

Like you said `inventory_hostname` is always defined (AFAICT) and is reliable, so let's just use it outright.

mike-nguyen: :thumbsup:

PR #45, opened 2016-08-25
improved-sanity-test: Add tests for atomic cli

mike-nguyen:
This brings in some simple atomic cli tests to the sanity suite. The
subcommands include pull, run, stop, install, uninstall, host status,
and scan.  The same set of tests are run pre and post upgrade.

The improved-sanity-tests playbook is taking longer to run due to
pull large container pulls and pulls from docker.io.  To reduce
overall runtime the atomic tests, the following choices were made:

- Use small rsyslog and busybox container for pull, install, uninstall
- Make atomic scan scan its own openscap image
- Leverage the docker container built in #42 by running atomic run
   and stop between the docker build and docker run commands
- Place cockpit_verify role earlier after the reboot to avoid pulling
   cockpit image twice


roles/atomic_installation_verify/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
+---
```
miabbott: If we are going to use the `rhel7/rsyslog` container, we should probably limit this to just RHEL hosts.

There are some Fedora Dockerfiles that have the `INSTALL` label, so perhaps we should be using those for the Fedora and CentOS cases.  (The CentOS Dockerfiles repo didn't appear to have anything lightweight for this purpose)

https://github.com/fedora-cloud/Fedora-Dockerfiles/search?utf8=%E2%9C%93&q=LABEL+INSTALL

roles/atomic_pull_verify/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    atomic_images: "atomic images list"
+  when: (atomic_minor.stdout|int) >= 12 and (atomic_major.stdout|int) == 1
+
+- name: Get atomic images
+  command: "{{ atomic_images }}"
+  register: aio
+
+- name: Fail if httpd is not in atomic images output
+  fail:
+    msg="httpd is not in atomic images output"
```
miabbott: The text here should be updated to refer to `busybox` rather than `httpd`

roles/atomic_scan_verify/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  shell: python -c "import pty; pty.spawn(['atomic', 'scan', '--scanner', 'openscap', 'rhel7/openscap'])"
+  register: atomic_output
+
+- name: Catch atomic scan failures
+  fail:
+    msg="Atomic scan failed"
+  when: "'Files associated' not in atomic_output.stdout"
+
+- name: Get scanner result directory
+  shell: echo "{{ atomic_output.stdout }}" | awk '/^Files associated/{print $8}' | sed 's/\.//'
```
miabbott: You should use the `quote` filter when passing variables to the `shell:` module

https://github.com/openshift/openshift-ansible/blob/master/docs/best_practices_guide.adoc#The-Ansible-quote-filter-MUST-be-used-with-any-variable-passed-into-the-shell-module

roles/atomic_scan_verify/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
+---
```
miabbott: This test is kind of tricky because I believe that OpenSCAP can only scan RHEL containers and eventually I could see a future where RHEL containers are available only via a subscription.

I guess we can run with this for now.  I wonder if there is any benefit to using the Fedora OpenSCAP image in the Fedora/CentOS case?

https://hub.docker.com/r/fedora/atomic_scan_openscap/

miabbott:
The `tests/improved-sanity-tests/info.txt` file should also be updated with the new tests that have been added.

miabbott:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;2LU4x7Rh&#x2F;355-3-expand-a-h-t-sanity-suite-to-include-atomic-cli-tests"> (3) Expand a-h-t sanity suite to include Atomic cli tests</a>

roles/atomic_scan_verify/tasks/main.yml
mike-nguyen: If that is the case, does it make sense to run this on anything other than RHELAH?  We can keep it as is and when RHEL containers are no longer available, I can add a when clause to the role.

cgwalters:
Took a quick look, seems sane to me.

miabbott:
LGTM

PR #47, opened 2016-09-22
Reset refspec in origin for alternate remotes

mike-nguyen:
The subscription manager changes the remote in the refspec in the
origin file.  If the test system is pointing to other update streams
like autobrew builds, the refspec is overwritten and causes issues
with upgrading.  To fix this, the refspec is now saved when it is not
the default value and re-written after the subscription process.

This fix is particularly important for tests that deal with building
RHEL containers and upgrading to alternate remotes.  Building RHEL
containers requires entitlements from the host.  Upgrades to alernate
remotes requires the proper refspec in the origin file.

### DIFF
--- a/roles/redhat_subscription/tasks/main.yaml
+++ b/roles/redhat_subscription/tasks/main.yaml
@@ -37,8 +37,8 @@
 
   - name: Set version and commit if deployment 1 is booted
     set_fact:
-      booted_commit: "{{ json['deployments'][0]['checksum'] }}"
-    when: json['deployments'][0]['booted']
+      booted_commit: "{{ json['deployments'][1]['checksum'] }}"
+    when: json['deployments'][1]['booted']
 
   - name: Get refspec from origin file
     command: grep refspec /ostree/deploy/rhel-atomic-host/deploy/{{ booted_commit }}.0.origin
--- a/roles/redhat_subscription/tasks/main.yaml
+++ b/roles/redhat_subscription/tasks/main.yaml
@@ -72,9 +72,13 @@
       --enable rhel-7-server-optional-rpms
       --enable rhel-7-server-extras-rpms
 
+  - name: Mask and stop rhsmcertd
+    shell: >
+      systemctl mask rhsmcertd && systemctl stop rhsmcertd
+
   - name: Revert refspec in origin after susbcription if it not the default refspec
     replace:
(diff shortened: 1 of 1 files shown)


cgwalters:
Isn't it sufficient to `systemctl mask rhsmcertd && systemctl stop rhsmcertd`?  I think in theory we'll lose certificate renewals but I can't imagine we'll hit that in short-running test cases.

mike-nguyen:
The refspec in the origin file is also changed during the subscription process with `systemctl mask rhsmcertd && systemctl stop rhsmcertd` so it will need to be changed back to the autobrew.  I noticed that after a reboot (with subscriptions and autobrew refspec), the refspec in the origin file is unchanged.  

The certCheckInterval in /etc/rhsm/rhsm.conf is set to 4 hours so our test should not be affected by this.  I will add `systemctl mask rhsmcertd && systemctl stop rhsmcertd` as an extra precaution to ensure the autobrew remote is not overwritten on the reboots after upgrade/rollback.

roles/redhat_subscription/tasks/main.yaml
```diff
@@ -21,6 +22,27 @@
...
+
+  - name: Set version and commit if deployment 0 is booted
+    set_fact:
+      booted_commit: "{{ json['deployments'][0]['checksum'] }}"
+    when: json['deployments'][0]['booted']
+
+  - name: Set version and commit if deployment 1 is booted
+    set_fact:
+      booted_commit: "{{ json['deployments'][0]['checksum'] }}"
+    when: json['deployments'][0]['booted']
```
miabbott: This looks like a copy/paste error.  I think you want to use the deployment at index 1.

miabbott:
> Ok, yeah. One side note, might be better to use the service module rather than systemctl directly, but I don't think it matters much here either.

@cgwalters I was going to suggest the same, but I don't think the service module supports masking

mike-nguyen:
> > Ok, yeah. One side note, might be better to use the service module rather than systemctl directly, but I don't think it matters much here either.

> @cgwalters  I was going to suggest the same, but I don't think the service module supports masking

service doesn't support masking.  There is a systemd module in Ansible 2.2 but these tests were developed and run on Ansible 2.1

mike-nguyen:
@miabbott please take another look when you get a chance.  I think I made all the fixes from your comments.

PR #48, opened 2016-09-22
docker_run_httpd: retry connection to httpd container

miabbott:
The internal CI job occasionally fails because the connectivity test
to the httpd container is unsuccessful. There does not seem to be a
good reason for this, so we'll just retry a few times.


mike-nguyen:
Looks good to me.

PR #50, opened 2016-09-28
Vagrantfile: adjust logic for CentOS

miabbott:
When trying to deploy HEAD^ for CentOS, I encountered an error because
the version of C7AH in the current Vagrant box is equivalent to HEAD^.

I adjusted the logic around the deploy to skip out if the version of
the Vagrant box happened to be HEAD^ from the remote.

Additionally, I had to use some `sed` edits to disable GPG
verification because of ostreedev/ostree#518


mike-nguyen:
LGTM.

PR #51, opened 2016-10-04
Atomic pull

mike-nguyen:
Use fully qualified container name for cockpit

There is an upstream issue #674 filed in the atomic CLI project for
the pull that occurs during atomic install.  The pull does not
resolve the fully qualified name for the container and throws an
error.  We should be using fully qualified container names so we
know we are pulling the right image from the right registry anyhow.


mike-nguyen:
Closing this and resubmitting a clean PR.

PR #53, opened 2016-10-06
atomic_installation_verify: use :latest tag

miabbott:
In projectatomic/skopeo#227, the `atomic install` command is failing
because there is no explicit tag specified.

This change makes the test use the `:latest` tag explicitly on the
cockpit images.

(We may want to revert this once the upstream issue is fixed)


miabbott:
@mike-nguyen please have a look

mike-nguyen:
LGTM.

PR #55, opened 2016-11-01
Use static hostname for 3-way merge check

mike-nguyen:
The ansible_hostname fact was changing after reboots in the test
environment causing the improved sanity tests to fail randomly.  A
similar issue occurred in issue #38 with the ansible_fqdn fact.
Using a static hostname still verifies the 3-way merge  after an
upgrade without dealing with issues related to the test environment.


miabbott:
Even though the host value is only used in two places, could we make it a global variable so we don't have to change things in two places?

mike-nguyen:
@miabbott Please take another look when you get the chance.  I changed the hostname into a variable.

miabbott:
:shipit:

PR #56, opened 2016-11-04
Move away from deprecated docker module

ashcrow:
The docker module is now considered deprecated. This change moves to
using the newer docker_container (2.1.0) and docker_image (1.5)
modules.

See: http://docs.ansible.com/ansible/docker_module.html#deprecated

Part of #28

PR #57, opened 2016-11-07
Revert "Move away from deprecated docker module (#56)"

mike-nguyen:
This reverts commit c9877e3b8f8ae902a0830fbcb4e5239dc6fb9752.
The tests are pinned to Ansible version 2.1 so the changes to
the new docker module do not work.


miabbott:
The Ansible docs indicate that the new `docker_container` and `docker_image` modules should be used with version 2.2.  Since we are currently using Ansible 2.1, let's just stick with the current approach.

PR #58, opened 2016-11-08
roles: check for presence of running httpd container

miabbott:
While debugging a recent failure on the internal autobrew stream, it
wasn't immediately obvious from the Ansible output that the httpd
container was not running.  I added a check after starting the
container to see if the container was in the output of `docker ps`.

Additionally, I opted to make use of a global variable in the httpd
image/container roles.  We previously took the `g_osname` variable and
concatenated some text onto it to make the values used in various
`docker` operations.


mike-nguyen:
LGTM and the playbook ran fine against RHELAH.

PR #59, opened 2016-11-08
tests: tag individual plays in sanity test

miabbott:
This adds tags on the higher order plays contained in the
`improved-sanity-test` playbook.  This allows us even more flexibility
in running specific tests when necessary.


mike-nguyen:
LGTM.  Sanity tested on RHEL Atomic.

PR #60, opened 2016-11-14
Fix apiserver for fully containerized kubernetes

mike-nguyen:
Atomic host now requires kubernetes to be run from a container because kubernetes
was removed from Atomic Host.  The apiserver-pod file was using
/etc/kubernetes/apiserver which did not work in the fully containerized
model of kubernetes.  The contents of the apiserver file was moved into the
apiserver-pod.json file and the edits on /etc/kubernetes/apiserver was
removed from the role.

### DIFF
--- a/roles/kubernetes_setup/tasks/main.yaml
+++ b/roles/kubernetes_setup/tasks/main.yaml
@@ -2,14 +2,14 @@
 #
 #  This role sets up kubernetes apiserver, controller-mgr, and scheduler pods
 #
-#  - name: pull {{ item }} 
+#  - name: pull {{ item }}
 #    command: docker pull registry.access.redhat.com/rhel7/{{ item }}
 #    with_items:
 #      - kubernetes-apiserver
 #      - kubernetes-controller-mgr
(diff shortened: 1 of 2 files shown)


miabbott:
Shouldn't this task also get removed as part of this PR?

https://github.com/projectatomic/atomic-host-tests/blob/master/roles/kubernetes_setup/tasks/main.yaml#L42

mike-nguyen:
@miabbott Isn't that removed?  I see it in the changes.

roles/kubernetes_setup/tasks/main.yaml
```diff
@@ -39,15 +39,9 @@
       regexp='^KUBELET_ARGS=\"\"'
       replace='KUBELET_ARGS=\"--register-node=true --config=/etc/kubernetes/manifests/\"'
 
-  - name: add kubelet args
```
mike-nguyen: @miabbott

miabbott:
Yup.  Not sure how I missed that.

PR #62, opened 2016-12-02
selinux_verify: add check for getfattr

miabbott:
We've run into situations where the `getfattr` utility is not
installed on the host and the failure message from this role is not
explicit in that case.  So I added a check for the utility ahead of
trying to use it.

I'm not going to bother trying to check SELinux information using
different methods, since this utility should be commonly installed on
AH and if it isn't, that is a problem to flag.


mike-nguyen:
Looks pretty straight forward

PR #63, opened 2016-12-06
Add package layering test

mike-nguyen:
Because roles and tasks cannot be be combined in Ansible, roles were created
for positive scenarios in package install and uninstall and its verifications
only.

Negative tests and tests with unexpected output are done afterwards in
the tasks portion of the playbook. Unfortunately there is a lot of repeated
code to validate output in the task section because we cannot combine roles.


tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+    - name: Fail if {{ g_pkg1 }} is installed
+      command: command -v {{ g_pkg1 }}
+      register: binary
+      failed_when: binary.rc != 1
+
+    # Verify that after additional repos are enabled (i.e. EPEL), rpm-ostree install|uninstall can be successfully used for packages in those repos.
+    - name: Retrieve EPEL rpm
+      get_url:
+        url: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+        dest: /tmp
```
cgwalters: I'd write to `/root`, otherwise it's a predictable filename in `/tmp`.

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+        msg: "{{ g_pkg1 }} not in rpm-ostree status output"
+      when: "'{{ g_pkg1 }}' not in pending_pkgs"
+
+    - name: Fail if {{ g_pkg1 }} is installed
+      command: command -v {{ g_pkg1 }}
+      register: binary
+      failed_when: binary.rc != 1
+
+    # Verify that after additional repos are enabled (i.e. EPEL), rpm-ostree install|uninstall can be successfully used for packages in those repos.
+    - name: Retrieve EPEL rpm
```
cgwalters: In somewhere else, I forget now, I just wrote the `.repo` and gpg key via `copy/contents` rather than fetching the RPM.  There's also various ansible modules for writing yum repo configs that would make sense.

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+        cpio -iv --to-stdout ./etc/yum.repos.d/epel.repo >
+        /etc/yum.repos.d/epel.repo
+
+    - name: Install {{ g_epel_pkg }} from EPEL
+      command: rpm-ostree install {{ g_epel_pkg }} -r
+      async: 60
+      poll: 0
+      ignore_errors: true
+
+    - name: wait for host to go down
```
cgwalters: Feels like we have a lot of copies of this, might make sense to have this be an include?

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+      shell: >
+        rpm2cpio /tmp/epel-release-latest-7.noarch.rpm |
+        cpio -iv --to-stdout ./etc/yum.repos.d/epel.repo >
+        /etc/yum.repos.d/epel.repo
+
+    - name: Install {{ g_epel_pkg }} from EPEL
+      command: rpm-ostree install {{ g_epel_pkg }} -r
+      async: 60
+      poll: 0
+      ignore_errors: true
```
cgwalters: Hm, but we don't want to ignore errors...let's do the reboot as a separate step?

cgwalters:
Nice work!

roles/rpm_ostree_install_verify/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+    installed_pkgs: "{{ installed_json['deployments'][1]['packages'] }}"
+  when: installed_json['deployments'][1] is defined and installed_json['deployments'][1]['booted']
+
+- name: Fail if {{ package }} is not in rpm-ostree status output
+  fail:
+    msg: "{{ package }} not in rpm-ostree status output"
+  when: "'{{ package }}' not in installed_pkgs"
+
+- name: Check for {{ package }} binary
+  command: command -v {{ package }}
```
jlebon: I would add `rpm -q` to the list of checks here, to make sure rpmdb was updated properly.

jlebon: Another cool test might be `rpm -ql` and then checking that all the files are present (maybe just modulo those in `/usr/share` since whatever tree you're testing was probably composed without docs).

roles/rpm_ostree_uninstall/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: Uninstall {{ packages }}
+  command: rpm-ostree uninstall {{ packages }}
+  when: not reboot
+
+- name: Uninstall {{ packages }} and reboot
+  command: rpm-ostree uninstall {{ packages }} -r
+  async: 60
+  poll: 0
+  when: reboot
+  ignore_errors: true
```
jlebon: Here too.

roles/rpm_ostree_uninstall_verify/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+    installed_json: "{{ installed.stdout | from_json }}"
+
+- name: Set packages when deployment 0 is booted
+  set_fact:
+    installed_pkgs: "{{ installed_json['deployments'][0]['packages'] }}"
+  when: installed_json['deployments'][0] is defined and installed_json['deployments'][0]['booted']
+
+- name: Set packages when deployment 1 is booted
+  set_fact:
+    installed_pkgs: "{{ installed_json['deployments'][1]['packages'] }}"
```
jlebon: But won't this fail if the `packages` entry is missing from the JSON, which is the case when no packages are layered?

jlebon: Maybe change these facts to just retrive the booted deployment entry and then just under make sure that there is no `packages` entry in the dict.

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+
+  roles:
+  # Subscribe if the system is RHEL
+    - role: redhat_subscription
+      when: ansible_distribution == 'RedHat'
+      tags:
+        - redhat_subscription
+
+    # single package install and uninstall
+    - { role: rpm_ostree_install, packages: "{{ g_pkg1 }}", reboot: false, tags: ['rpm_ostree_install11'] }
```
jlebon: Any reason for not unpacking these dicts on multiple lines to make it more legible?

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+
+    - { role: rpm_ostree_uninstall_verify, package: "{{ g_pkg1 }}", tags: ['rpm_ostree_uninstall_verify14'] }
+
+
+    # multiple package install and uninstall
+    - { role: rpm_ostree_install, packages: "{{ g_pkg1 }} {{ g_pkg2 }}", reboot: false, tags: ['rpm_ostree_install21'] }
+
+    - role: reboot
+      tags:
+        - reboot3
```
jlebon: Ahh OK, I see what you were going for with the reboot flag now. In reality, the `-r` flag just sums up to us calling `systemctl reboot` after the transaction finishes. I'd say it's probably not worth rerunning every test both with and without `-r`.

tests/pkg-layering/main.yml
mike-nguyen: I'll change them over to multiple lines

tests/pkg-layering/main.yml
mike-nguyen: Since roles and tasks can't be mixed in a playbook, I wrote the role to accept a reboot parameter.  That way I can test the scenario with and without the reboot flag in one role.

To more clearly test package layering, I separated the install and reboot for anything that is not explicitly a reboot flag test.

tests/pkg-layering/main.yml
mike-nguyen: I haven't used includes before but it looks like the way to go.  Thanks for the tip!

tests/pkg-layering/main.yml
miabbott: Perhaps retrieving them from cgit - http://pkgs.fedoraproject.org/cgit/rpms/epel-release.git/tree/?h=epel7

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+#
+# Negative Testing
+#   - Verify correct error is displayed removing a package that does not exist
+#   - Verify correct error is displayed when adding a package that does not exists in the repo
+#   - Verify correct error message is displayed when adding a package that is already in the deployment
+#   - Verify packages which have files with non-root ownership cannot be installed
+#   - Verify that install/uninstall operations cannot be performed by unprivileged users
+#
+
+- name: Package Layering Tests
```
miabbott: Maybe this test should be split into multiple plays, one per test.  I think it would solve the issue of using unique tags (for the most part) and would improve readbility of this playbook.

tests/pkg-layering/main.yml
```diff
@@ -0,0 +1,319 @@
...
+    - name: Set packages when deployment 0 is booted
+      set_fact:
+        installed_pkgs: "{{ installed_json['deployments'][0]['packages'] }}"
+      when: installed_json['deployments'][0] is defined and installed_json['deployments'][0]['booted']
+
+    - name: Set packages when deployment 1 is booted
+      set_fact:
+        installed_pkgs: "{{ installed_json['deployments'][1]['packages'] }}"
+      when: installed_json['deployments'][1] is defined and installed_json['deployments'][1]['booted']
+
```
miabbott: Perhaps this status checking could also be moved into an include?

tests/pkg-layering/main.yml
mike-nguyen: I ended up adding the repo and writing the GPG key manually.  I see advantages and disadvantages in both approaches.  

If we grab the gpg key and repo file from cgit or from the rpm, we always have the latest file but our test will fail if there are issues with the web server .  On the flip side, if we manually write the repo file and GPGkey, we  don't rely on the webserver but will fail if the gpg key or the repo URL changes.

roles/rpm_ostree_install_verify/tasks/main.yml
mike-nguyen: added the rpm -q check.  I will looking into the -ql file check!

mike-nguyen: @jlebon I'm looking at rpm -ql.  Are we supposed to installing the docs with pkg layering or not?  I know the packages that are installed at compose time don't have the docs installed but layered packages have the docs.  I can basicallly run rpm -ql and check that every file in there exists.

cgwalters: In the future we may change layering to honor the host configured options around documentation/translations.  So, don't test every file right now, I'd just test `/usr/bin/foo` for package `foo` for a few key cases etc.

mike-nguyen: I'll leave this test out for now then and just check rpm -q and that the binary exists

tests/pkg-layering/main.yml
mike-nguyen: All tests are separated into plays now.  It should also help with helping understand the test output.

tests/pkg-layering/main.yml
mike-nguyen: I moved these into an include.  It is basically a duplication of the verification roles.

miabbott:
I ran this playbook against CAHC, F24, F25, and RHEL AH.  Everything worked as expected.  Great work! 👌

PR #64, opened 2016-12-08
Vagrantfile: add F25, remove F23, update CAHC

miabbott:
This removes the F23 testing option, adds the F25 testing option, and
changes CAHC to pull the Vagrant box from the ci.centos.org location.


cgwalters:
LGTM.

PR #65, opened 2016-12-08
roles: update SELinux contexts for F25

miabbott:
Turns out the SELinux policy on F25 has `svirt_sandbox_file_t` is an
alias to `container_file_t`

```
$ seinfo -x --type=svirt_sandbox_file_t
   TypeName container_file_t
   Aliases
      svirt_sandbox_file_t
      svirt_lxc_file_t
```

This was causing some of the SELinux checks to fail on F25, so now they are
conditional on F25.

### DIFF
--- a/roles/var_add_files/tasks/main.yml
+++ b/roles/var_add_files/tasks/main.yml
@@ -1,6 +1,17 @@
 ---
 # vim: set ft=ansible:
 #
+- name: Set fedora25 to be False by default
+  set_fact:
+    fedora25: False
+
+- name: Set fedora25 to be True for F25
+  set_fact:
+    fedora25: True
+  when:
+    - ansible_distribution == "Fedora"
+    - ansible_distribution_version == "25"
+
 - name: Fail if vaf_filename is not defined
   fail:
     msg: "vaf_filename is undefined"
@@ -11,12 +22,21 @@
     msg: "vaf_dirname is undefined"
   when: vaf_dirname is undefined
 
-- name: Make directory in /var
+- name: Make directory in /var with svirt_sandbox_file_t
   file:
     path: "/var/{{ vaf_dirname }}"
     state: directory
     mode: 0755
     setype: svirt_sandbox_file_t
+  when: not fedora25
+
+- name: Make directory in /var with container_file_t
+  file:
+    path: "/var/{{ vaf_dirname }}"
+    state: directory
+    mode: 0755
+    setype: container_file_t
+  when: fedora25
 
 - name: Add file in new directory
   shell: "echo $(sha256sum /etc/passwd) > {{ vaf_filename | quote }}"
--- a/roles/var_add_files/tasks/main.yml
+++ b/roles/var_add_files/tasks/main.yml
@@ -1,16 +1,19 @@
 ---
 # vim: set ft=ansible:
 #
-- name: Set fedora25 to be False by default
+- name: Set use_container_file_t to be False by default
   set_fact:
-    fedora25: False
+    use_container_file_t: False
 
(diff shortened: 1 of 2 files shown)


roles/var_add_files/tasks/main.yml
```diff
@@ -1,6 +1,17 @@
...
+- name: Set fedora25 to be False by default
+  set_fact:
+    fedora25: False
+
+- name: Set fedora25 to be True for F25
+  set_fact:
+    fedora25: True
+  when:
+    - ansible_distribution == "Fedora"
+    - ansible_distribution_version == "25"
```
cgwalters: `>=` ?

cgwalters: (And if we do this, the fact name is like `have_new_container_sepolicy` or something)?

cgwalters: Given it's highly likely at some point this gets downstream, I'd try to detect it - something like run `chcon -t container_file_t foo` and see whether or not that works.

roles/var_files_present/tasks/main.yml
```diff
@@ -54,5 +70,10 @@
     msg: "The permissions on the added file are incorrect"
   when: added_file.stat.mode != "0644"
 
-- name: Check if the SELinux context on the added file is correct
+- name: Check for svirt_sandbox_file_t context on added file
   shell: "stat /var/{{ vfp_dirname | quote }}/{{ vfp_filename | quote}} | grep svirt_sandbox_file_t"
```
cgwalters: You can use `stat -c '%C' /path/to/file` to avoid the grep, and that would make it easier to register it as a variable, then see whether or not it's in a set/list.

mike-nguyen:
Does it make sense to register the output of  `seinfo -x --type=svirt_sandbox_file_t` then grab the selinux context of the file using `stat -c '%C' /path/to/file`  and fail when the selinux context is not in the seinfo output?  

The advantages I see is that we don't have to do any version checks and sets and will work for any aliases.  The disadvantage is that is not as precise.

miabbott:
Thanks for the comments so far.  I'm working on some changes to address the ideas raised.

miabbott:
Changed the approach to detect the `container_file_t` type and used `stat -c '%C'` instead of `grep`

roles/var_add_files/tasks/main.yml
```diff
@@ -1,16 +1,19 @@
...
+  ignore_errors: True
+
+- name: Set use_container_file_t to True when present
   set_fact:
-    fedora25: True
-  when:
-    - ansible_distribution == "Fedora"
-    - ansible_distribution_version == "25"
+    use_container_file_t: True
+  when: seinfo.rc == 0
```
jlebon: Hmm, I wonder if you can simplify this even more. E.g.

```yaml
set_fact:
  container_label: '{{ "container_file_t" if seinfo.rc == 0 else "svirt_sandbox_file_t" }}'
```

And then lower down, just have a single test that does:

```yaml
failed_when: "container_label not in stat.stdout"
```

miabbott: Efficient!  I will investigate this path.

mike-nguyen: Storing this gold away in my memory.

miabbott:
Big up to @jlebon and his suggestion! 🙌  I also fixed another instance of 'stat` that I missed originally.

miabbott:
In a future PR, I think we should move some of this `set_fact` business into a separate role.

mike-nguyen:
So fresh and so clean now :+1:

PR #66, opened 2016-12-12
atomic_scan_verify: use upstream scanner

miabbott:
We were seeing failures on certain streams when trying to install the
`openscap` scanner from `registry.access.redhat.com`.  The version of
`skopeo` being used in these cases did not support v1 schema manifests
and that is the only version available on that registry.

Using the scanner from upstream alleviates this problem and gives us
further coverage in this space.

Oddly, only RHEL-based images are supported by the `openscap` scanner,
so the behavior has been changed to scan the RHEL7 image regardless if
the scanner used is upstream or downstream.

Finally, `atomic scan` no longer requires a TTY, so we can change
how that command is invoked.

### DIFF
--- a/roles/atomic_scan_verify/tasks/main.yml
+++ b/roles/atomic_scan_verify/tasks/main.yml
@@ -49,7 +49,7 @@
   when: "'Files associated' not in atomic_output.stdout"
 
 - name: Get scanner result directory
-  shell: echo "{{ atomic_output.stdout | quote }}" | awk '/Files associated/{print $8}' | sed "s/\.[\']*//"
+  shell: echo {{ atomic_output.stdout | quote }} | awk '/^Files associated/{print $8}' | sed "s/\.[\']*//"
   register: dir
 
 - name: Verify scanner result directory exists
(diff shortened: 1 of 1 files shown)


roles/atomic_scan_verify/tasks/main.yml
```diff
@@ -9,18 +9,36 @@
...
+  command: atomic install {{ scanner_image }}
   register: result
 
 - name: Check installation of openscap
   fail:
     msg="Atomic install was unsuccessful"
   when: "'Installation complete' not in result.stdout"
 
+- name: Pull scanner target
+  command: atomic pull {{ scanner_target }}
```
mike-nguyen: Unfortunately, the same skopeo problem pulling from registry.access.redhat.com also exists pulling the rhel7 container

mike-nguyen: Can we pull a rhel7 image from somewhere else?  I know that is not ideal but little else we can do for now.

miabbott: Hmm...I didn't encounter that in my testing.  Maybe we could just use `docker pull` here.

miabbott:
Yeah, but we don't want to unquote variables when using the `shell` module.

I'll tinker with the `awk` usage and see if I can get it to work.

miabbott:
Removed the quotes and tweaked the `awk` statement slightly.  Should now work if `--debug` is specified or not.

mike-nguyen:
Looks good! :+1:

PR #67, opened 2016-12-12
Add semanage add test

mike-nguyen:
Test to make sure that policies can be added to set selinux context
of files within a directory.

A regression test for BZ1398427.

### DIFF
--- a/roles/semanage_add_verify/tasks/main.yml
+++ b/roles/semanage_add_verify/tasks/main.yml
@@ -1,0 +1,22 @@
+---
+- name: Check for presence of container_file_t type
+  command: seinfo --type=container_file_t
+  register: seinfo
+  ignore_errors: True
+
+- name: Set SELinux type to be used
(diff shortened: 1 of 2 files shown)


roles/semanage_add_verify/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Check for presence of container_file_t type
+  command: seinfo --type=container_file_t
+  register: seinfo
+  ignore_errors: True
+
+- name: Set SELinux type to be used
+  set_fact:
+    selinux_type: '{{ "container_file_t" if seinfo.rc == 0 else "svirt_sandbox_file_t" }}'
+    dir_match: '/var/lib/kubelet(/.*)?'
```
miabbott: I would like to move these types of global `set_fact` tasks into one role we can run at the beginning of each playbook.

Not for this review, but something we should do in the near future.

PR #69, opened 2016-12-19
Add admin unlock tests

mike-nguyen:
Adding ostree admin unlock tests.  The strategy to testing admin
unlock is to install packages from the EPEL repo.  I scraped the
HTML for the package named under the assumption that the packages
could and get a new name with the version.

Roles were used for most of the positive testing but negative testing
was done as tasks.  Since roles and tasks cannot be combined, include
was used to re-use roles or to uses roles within roles.


roles/rpm_install/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+  set_fact:
+    epel_base: "https://dl.fedoraproject.org/pub/epel/7/x86_64"
+
+- name: Get first letter of package name
+  shell: echo {{ rpm_name | quote }} | head -c 1
+  register: first_letter
+
+- name: Get base url
+  set_fact:
+    epel_url: "{{ epel_base }}/{{ first_letter.stdout }}"
```
miabbott: You can use common Python string operations with variables like this:

`epel_url: "{{ epel_base }}/{{ rpm_name[0] }}"`

This would eliminate a task there.

tests/admin-unlock/main.yml
```diff
@@ -0,0 +1,316 @@
...
+  tags:
+    - unlock_add_remove_package_non_persistence
+
+  vars_files:
+    - vars.yml
+
+  pre_tasks:
+    - name: set rpm_name
+      set_fact:
+        rpm_name: "{{ g_rpm_name }}"
```
miabbott: Rather than setting the `rpm_name` here, perhaps pass it in as an argument for role that requires it?

Since that method is used when installing multiple RPMs, it might make sense to use that method everywhere.

tests/admin-unlock/main.yml
```diff
@@ -0,0 +1,316 @@
...
+    - vars.yml
+
+  tasks:
+    - name: set rpm_name
+      set_fact:
+        rpm_name: "{{ g_rpm_name }}"
+
+    - name: Get one of the rpms that was downloaded
+      shell: ls /root | grep {{ rpm_name }}*
+      register: rpm
```
miabbott: So this assumes that one of the previous plays ran on the host and downloaded an RPM, correct?

I don't really like making that assumption, but the alternative is duplicating a bunch of tasks here.  😞

tests/admin-unlock/main.yml
```diff
@@ -0,0 +1,316 @@
...
+  vars_files:
+    - vars.yml
+
+  tasks:
+    - name: Unlock system
+      command: ostree admin unlock
+
+    - include: 'roles/rpm_ostree_status/tasks/main.yml'
+
+    - name: Verify rpm_ostree status has unlocked set to hotfix
```
miabbott: I think this should read:  `Verify rpm_ostree status is unlocked`

tests/admin-unlock/main.yml
```diff
@@ -0,0 +1,316 @@
...
+      fail:
+        msg: "Booted deployment does not have unlocked set to hotfix"
```
miabbott: Reword as:  `Booted deployment is not unlocked`

roles/rpm_uninstall/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+#
+# Removes package using rpm
+#
+- name: Fail if g_rpm_name
+  fail:
+    msg: "g_rpm_name is undefined"
+  when: g_rpm_name is undefined
+
+- name: Install {{ g_rpm_name }}
+  command:  rpm -e {{ g_rpm_name }}
```
miabbott: I feel like this role should also be using `rpm_name` as the variable.  I know it is not currently used with a different value than `g_rpm_name` but we should provide that flexibility.

tests/admin-unlock/main.yml
mike-nguyen: That's an assumption that is made because we don't technically restore to a clean state between each test. I was under the impression that separating the plays was to produce easier to read test logs not necessarily to separate the tests.  Do you think it is better to duplicate a bunch of tasks or just add this into the README.md?

tests/admin-unlock/main.yml
mike-nguyen: I changed it to "verify rpm_ostree status is unlocked and set to hotfix".  I think that should clear things up as it can be in three states--unlocked: hotfix, unlocked: development, and unlocked: none.

tests/admin-unlock/main.yml
mike-nguyen: Changed to: "Booted deployment is not unlocked and set to hotfix"

roles/rpm_install/tasks/main.yml
mike-nguyen: Awesome, I'll use that!

roles/rpm_uninstall/tasks/main.yml
mike-nguyen: I have no idea why I did this for one role but not the other :disappointed:

tests/admin-unlock/main.yml
miabbott: I'd opt for less duplication, so we'll leave it as is.  A note in the README would be good.

roles/ostree_admin_unlock_hotfix/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+# vim: set ft=ansible:
+#
+# Unlocks the system to be unlocked to make temporary changes to the /usr mount
```
jlebon: s/temporary/permanent/ ?

roles/overlayfs_verify_missing/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# vim: set ft=ansible:
+#
+# Runs the mount command and verifies no overlayfs is present
+#
+- name: Verify overlayfs
+  shell: mount | grep -e 'overlay on .* type overlay'
```
jlebon: Hmm, I would also check for e.g. "ostree" in that regex.

roles/overlayfs_verify_present/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+# vim: set ft=ansible:
+#
+# Verifies the presence of an overlayfs in mount output
+#
+- name: Verify overlayfs
+  shell: mount | grep -e 'overlay on .* type overlay'
```
jlebon: Ditto

roles/package_verify_missing/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+#
+- name: Fail if rpm_name is undefined
+  fail:
+    msg: "rpm_name must be passed into this role/include"
+  when: rpm_name is undefined
+
+- name: Check for {{ rpm_name }} binary
+  command: command -v {{ rpm_name }}
+  register: c_output
+  failed_when: c_output.rc != 1
```
jlebon: I'd do the negative here and below, e.g. `== 0`

roles/rpm_uninstall/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+# vim: set ft=ansible:
+#
+# Removes package using rpm
+#
+- name: Fail if rpm_name
+  fail:
+    msg: "rpm_name is undefined"
+  when: rpm_name is undefined
+
+- name: Install {{ rpm_name }}
```
jlebon: s/Install/Uninstall/ ?

tests/admin-unlock/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+With the plugin installed, you should be able to choose a CentOS AH box, a
+Fedora 23 AH box, or a CAHC box.
+
+```
+$ vagrant up centos
+
+or
+
+$ vagrant up fedora23
```
jlebon: 24 or 25 I'd say. 23 has just passed EOL.

tests/admin-unlock/main.yml
```diff
@@ -0,0 +1,313 @@
...
+
+  tags:
+    - unlock_add_remove_package_non_persistence
+
+  vars_files:
+    - vars.yml
+
+  roles:
+    - role: ostree_admin_unlock
+      rpm_name: "{{ g_rpm_name }}"
```
jlebon: Wait, why do you need to pass `rpm_name` here?

mike-nguyen: We don't.

roles/package_verify_missing/tasks/main.yml
mike-nguyen: Is there a particular reason why?  Should I swap over all my != 1 to ==0?

mike-nguyen:
@jlebon we're planning to a quick admin unlock test into the improved_sanity_test

roles/package_verify_missing/tasks/main.yml
jlebon: It's the difference between "I expect this command to fail and return code 1" vs. just "I expect this command to fail". I.e. we don't care here what error code it actually returns, we just care that it fails. (In reality, those particular commands likely won't change their error code, but in the general case it's just more bulletproof to check for zero/non-zero).

mike-nguyen: Ah, got it!

roles/overlayfs_verify_missing/tasks/main.yml
mike-nguyen: ostree doesn't appear in the regex if you are using ostree admin unlock --hotfix

I can make it check for ostree in the regex when it is not hotfix and not look for the ostree if it is hotfix or I can leave it as it is.  Any thoughts?

mike-nguyen: Eloquent solution!  I think that's all we really need.

mike-nguyen:
I think I have all the requested changes in.  Let me know if it looks good!

miabbott:
New changes LGTM.  Unless @jlebon has additional comments, I'm ready to merge this.

jlebon:
👍

PR #70, opened 2017-01-03
atomic_verify_install: move to non-interactive (un)installs

miabbott:
With the next release (version 127) of the `cockpit-ws` container,
non-interactive (un)installs will be possible as the Dockerfile has
been updated to not use the `-ti` flags.

This change brings support for non-interactive (un)installs by
checking the version of the `cockpit-ws` RPM and proceeding accordingly.

### DIFF
--- a/roles/atomic_installation_verify/tasks/main.yml
+++ b/roles/atomic_installation_verify/tasks/main.yml
@@ -22,17 +22,29 @@
     cockpit_cname: "docker.io/cockpit/ws:latest"
   when: ('CentOS' in ansible_distribution) or ansible_distribution == "Fedora"
 
+- name: Pull the cockpit container for inspection
+  command: docker pull {{ cockpit_cname }}
+
+- name: Get the version of the cockpit-ws rpm in the container
+  command: docker run --rm {{ cockpit_cname }} rpm --queryformat '%{VERSION}\n' -q cockpit-ws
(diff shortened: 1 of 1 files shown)


roles/atomic_installation_verify/tasks/main.yml
```diff
@@ -22,17 +22,29 @@
     cockpit_cname: "docker.io/cockpit/ws:latest"
   when: ('CentOS' in ansible_distribution) or ansible_distribution == "Fedora"
 
+- name: Pull the cockpit container for inspection
+  command: docker pull {{ cockpit_cname }}
+
+- name: Get the version of the cockpit-ws rpm in the container
+  command: docker run --rm {{ cockpit_cname }} rpm --queryformat '%{VERSION}\n' -q cockpit-ws
```
jlebon: Yuck. This would be better done by just embedding the version in a label instead.

Actually, that's the case in the RHEL version of the container (it has a `version` label). So maybe cockpit just has to upstream that bit.

jlebon: I opened https://github.com/cockpit-project/cockpit-container/issues/5

mike-nguyen:
I am okay with putting this in until the cockpit issue gets resolved then we can switch to using the label.

PR #71, opened 2017-01-06
Fix handling of systems (CAHC) with no builtin kube

cgwalters:
The previous change to test semanage for /var/lib/kubelet fails if the host
doesn't have kubelet installed, as is the case for CAHC.

Add a global `have_builtin_kube` fact that we can use elsewhere too.


cgwalters:
I didn't actually test this locally yet FWIW.

PR #73, opened 2017-01-09
Add system containers test

mike-nguyen:
This commit adds system container tests for atomic hosts.  Since
system containers are still relatively new and there are not many
system containers yet, the tests use gscrivano's hello-world container
to test system container functionality specified in the README.md file.

There is also a basic test for flannel and etcd system containers.
The longer term plan is to roll the flannel and etcd system containers
into the single note k8-cluster test for full functional testing.


tests/system-containers/main.yml
```diff
@@ -0,0 +1,511 @@
...
+        flannel_name: 'flannel'
+      when: ansible_distribution == 'RedHat'
+
+    - name: Set facts
+      set_fact:
+        etcd_image: 'gscrivano/etcd'
+        etcd_name: 'etcd'
+        flannel_image: 'gscrivano/flannel'
+        flannel_name: 'flannel'
+      when: ansible_distribution != 'RedHat'
```
miabbott: It looks like we may have proper upstream images soon.

https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2017-January/msg00024.html

When they land, we'll want to update this playbook to use them.

tests/system-containers/main.yml
```diff
@@ -0,0 +1,511 @@
...
+  #    - Verify error message is displayed when uninstalling a system container that does not exist
+  #    - Verify error message is displayed when installing a non system container with the --system option?
+
+  tags:
+    - negative
+
+  vars_files:
+    - vars.yml
+
+  tasks:
```
miabbott: How about a negative test that tries to install a container that does not exist?

Or what about trying to install a container that is not a system container?

tests/system-containers/main.yml
mike-nguyen: :+1:

tests/system-containers/main.yml
mike-nguyen: Let me add some more negative tests into this before you try running it yourself.

mike-nguyen:
Added a few more negative tests.  PTAL when you have the chance.

tests/system-containers/main.yml
mike-nguyen: The system containers code will try to make non system containers into system containers by creating the the files necessary to start the services.  It doesn't mean that it will work though.  I tried running atomic install --system with the rhel7 container and it created a rhel7 service.

miabbott:
[(3) [a-h-t] Create automated test for system containers](https://trello.com/c/kLmrhhTp/614-3-a-h-t-create-automated-test-for-system-containers)

miabbott:
Ran this against CAHC, F25AH, and RHELAH without issue.  Great work!

PR #75, opened 2017-01-10
docker_build_httpd:  add makecache.sh script

miabbott:
We have seen failures in the tests that appear to be related to
network issues (either in our OpenStack infra or in the Fedora infra)
when trying to install `httpd` while building the Docker image.

To combat this, we just steal the [env_make_cache()](https://github.com/jlebon/redhat-ci/blob/master/testrunner#L267)
routine from @jlebon and stick it in a dumb script.  Then we `ADD` to
the Docker image and `RUN` it before trying to install any packages.

This does add some extra instructions and time while building the
Docker image, but should help us avoid these network related failures.

### DIFF
--- a/roles/docker_build_httpd/files/centos_httpd_Dockerfile
+++ b/roles/docker_build_httpd/files/centos_httpd_Dockerfile
@@ -7,9 +7,9 @@ LABEL RUN="docker run -d --name NAME -p 80:80 IMAGE"
 ENV container docker
 
 ADD makecache.sh /
-RUN /makecache.sh
 
-RUN yum -y install httpd && \
+RUN /makecache.sh && \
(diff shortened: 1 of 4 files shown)


jlebon:
👍 

Pretty nit here given the purpose is to test things, but a better pattern to minimize total container size might be to incorporate it into the `RUN yum` step. Otherwise that `yum clean all` won't do much good. E.g.

```
 RUN /makecache.sh && \
     yum -y install httpd && \
     yum clean all
```

miabbott:
@jlebon I actually had that same thought after I submitted it.  I think I'll push a fixup for thtat.

jlebon:
👍

PR #77, opened 2017-01-11
docker_build_httpd: copy makecache.sh script

miabbott:
While the inclusion of the `makecache.sh` script is nice, it would
help if we actually copied it over to the host under test.

I used a separate `copy` statement to give the script executable
permissions, which the Dockerfile does not require.

### DIFF
--- a/roles/docker_build_httpd/tasks/main.yml
+++ b/roles/docker_build_httpd/tasks/main.yml
@@ -27,6 +27,14 @@
     group: root
     mode: 0644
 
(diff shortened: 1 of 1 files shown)


miabbott:
@jlebon @mike-nguyen please have a look

mike-nguyen:
LGTM

jlebon:
👍

mike-nguyen:
Quick test run before I merge.  Stay tuned.

PR #79, opened 2017-01-18
Add docker(-latest) tests

mike-nguyen:
This playbook adds some basic functionality tests for docker and
docker-latest.  It leverages much of the work that was done for
docker in the improved sanity tests.  There was one small change
to docker_run_httpd where the httpd container is started and
stopped again to provide coverage for the docker start command.

By default, the tests will configure the atomic host to run
docker-latest then run the suite of test.  There is a boolean
in vars.yml named g_docker_latest that can be set to false if
the tester wants to run the tests against the default docker
instance.


miabbott:
[(5) [a-h-t] Create a basic test suite for docker-latest functionality](https://trello.com/c/LtidYoHE/629-5-a-h-t-create-a-basic-test-suite-for-docker-latest-functionality)

roles/docker_latest_setup/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+# vim: set ft=ansible:
+#
+# This role configures the atomic host to run docker lates
+#
+
```
miabbott: There should be a check at the top of this role to see if `docker-latest` is installed.

I just found that F25AH doesn't not include the package, so the playbook bombed out.

mike-nguyen: Test will now fail with message if docker-latest is not installed.

miabbott:
Great work!  It actually already caught some breakage in CAHC when I ran it.  See CentOS/sig-atomic-buildscripts#233

mike-nguyen:
:+1:

tests/docker/main.yml
```diff
@@ -0,0 +1,100 @@
...
+
+  pre_tasks:
+    - name: Check for docker-latest package
+      command: rpm -qa
+      register: pkgs
+
+    - name: Fail if docker latest is not installed
+      fail:
+        msg: "This test is not valid because docker-latest is not installed"
+      when: "'docker-latest' not in pkgs.stdout"
```
miabbott: These pre-tasks need to be conditional to `g_docker_latest: true`

mike-nguyen: Good catch!  I totally  missed that.  Hopefully it should be good now.  I ran a quick test and it seems to logically work.

miabbott:
Keeping it 💯

PR #81, opened 2017-01-21
users: verify password hashing algorithm

miabbott:
CentOS, Fedora, and RHEL Atomic Host all conveniently use SHA-512 to
hash their passwords.

These changes verify that the hashing algorithm is correctly
configured to SHA-512 and also verifies that hashed passwords are
correctly stored in `/etc/shadow`

The password is randomly generated via `uuidgen`, then fed into
`passwd`.  The hash is confirmed by searching for the user and the
leading `$6` on the password hash in `/etc/shadow`.

See `man crypt` for details about the hashing algorithms.

### DIFF
--- a/roles/user_verify_present/tasks/main.yml
+++ b/roles/user_verify_present/tasks/main.yml
@@ -10,4 +10,4 @@
   command: "grep atomic-user-{{ uvp_uid }}:x:{{ uvp_uid }}:{{ uvp_uid }}::/home/atomic-user-{{ uvp_uid }}:/bin/bash /etc/passwd"
 
 - name: Verify password is a SHA-512 hash
-  command: "grep '^atomic-user-{{ uvp_uid }}:$6.*' /etc/shadow"
+  command: "grep '^atomic-user-{{ uvp_uid }}:$6$.*' /etc/shadow"
--- a/roles/user_verify_present/tasks/main.yml
+++ b/roles/user_verify_present/tasks/main.yml
@@ -8,3 +8,6 @@
 
 - name: Verify users in /etc/passwd
   command: "grep atomic-user-{{ uvp_uid }}:x:{{ uvp_uid }}:{{ uvp_uid }}::/home/atomic-user-{{ uvp_uid }}:/bin/bash /etc/passwd"
+
+- name: Verify password is a SHA-512 hash
+  command: "grep '^atomic-user-{{ uvp_uid }}:$6.*' /etc/shadow"
--- a/roles/user_add/tasks/main.yml
+++ b/roles/user_add/tasks/main.yml
@@ -6,6 +6,22 @@
     msg: "ua_uid variable is not defined"
   when: ua_uid is undefined
 
+- name: Check enabled password algorithm
+  shell: authconfig --test | grep hash
+  register: algo
+
+- name: Fail if password algorithm is not SHA-512
+  fail:
+    msg: "Password algorithm is not SHA-512"
+  when: "'sha512' not in algo.stdout"
+
 - name: Add user
   command: "useradd atomic-user-{{ ua_uid }} -u {{ ua_uid }}"
 
+  # because UUIDs could be passwords
+- name: Generate some random string
+  command: uuidgen -r
+  register: uuid
+
+- name: Set the password for the user
+  shell: "echo {{ uuid.stdout | quote }} | passwd atomic-user-{{ ua_uid | quote }} --stdin"


roles/user_verify_present/tasks/main.yml
```diff
@@ -8,3 +8,6 @@
 
 - name: Verify users in /etc/passwd
   command: "grep atomic-user-{{ uvp_uid }}:x:{{ uvp_uid }}:{{ uvp_uid }}::/home/atomic-user-{{ uvp_uid }}:/bin/bash /etc/passwd"
+
+- name: Verify password is a SHA-512 hash
+  command: "grep '^atomic-user-{{ uvp_uid }}:$6.*' /etc/shadow"
```
jlebon: Maybe use `$6$.*` instead of `$6.*`.

jlebon:
This is a great test!

roles/user_verify_present/tasks/main.yml
miabbott: Yeah, that would be correct.  Guess I should have read `man crypt` more carefully.

jlebon:
👍

mike-nguyen:
Code looks good to me.  It ran fine on RHEL AH and Fedora AH, but I ran into an issue running on CentOS AH.  It looks like hashing is md5sum instead of sha512.

`authconfig --test | grep hash`
`password hashing algorithm is md5`

`rpm-ostree status`
`State: idle`
`Deployments:`
`● centos-atomic-host:centos-atomic-host/7/x86_64/standard`
`       Version: 7.20170117 (2017-01-17 17:45:04)`
`        Commit: 5d8b88fbdde5e1df9f50a837f881da9d6031ba5a3af5f1681b9d38793a7f503f`
`        OSName: centos-atomic-host`
`  GPGSignature: 1 signature`
`                Signature made Tue 17 Jan 2017 06:00:06 PM UTC using RSA key ID F17E745691BA8335`
`                Good signature from "CentOS Atomic SIG <security@centos.org>"`
`      Unlocked: hotfix`

cgwalters:
Is that an old upgraded host?  This should have been fixed by https://github.com/CentOS/sig-atomic-buildscripts/commit/5f68d79ddb95d17d36e24b2dc8e25aecfd259d1e

mike-nguyen:
@cgwalters yes it is an old upgrade host.  Let me try with a newer CentOS AH host.  I was not running against CAHC.

mike-nguyen:
@cgwalters the newest CentOS Atomic Host (not continuous) is from 10/05 and has md5 hashing.  When I rebase from that to CAHC, the md5 hashing stays the same.  

If I boot up Fedora 25 AH and rebase to CAHC, I get the correct sha512 hashing.

mike-nguyen:
@miabbott In regards to this test suite, it looks like it runs fine against CAHC, Fedora, and RHELAH as long as the base image we're upgrading from had sha512 hashing.

miabbott:
@mike-nguyen Yeah, the CentOS AH case looks scary.  I was only testing on CAHC.  Even the most recent CentOS AH release from 1/17 uses MD5.  We'll have to file an bug for that.

```
# atomic host upgrade
Updating from: centos-atomic-host:centos-atomic-host/7/x86_64/standard
1 metadata, 0 content objects fetched; 313 B transferred in 0 seconds
No upgrade available.
# atomic host status
State: idle
Deployments:
● centos-atomic-host:centos-atomic-host/7/x86_64/standard
       Version: 7.20170117 (2017-01-17 17:45:04)
        Commit: 5d8b88fbdde5e1df9f50a837f881da9d6031ba5a3af5f1681b9d38793a7f503f
        OSName: centos-atomic-host

  centos-atomic-host:centos-atomic-host/7/x86_64/standard
       Version: 7.20161104 (2016-11-04 22:53:08)
        Commit: daf53c718bd1cd9b84b1b5d21843cd9b981d3634c2444c2bcb09cccee771058b
        OSName: centos-atomic-host
# authconfig --test | grep hashing
 password hashing algorithm is md5
```

I'm hesitant to change the test, since it actually seems to be doing what it was intended.  :)

What do others think?

miabbott:
https://bugs.centos.org/view.php?id=12720

mike-nguyen:
@miabbott  Thanks!  I was trying to figure out where to file that bug.  In terms of the test, I agree that it doesn't need to be changed.  The available CentOS AH image is pretty old and I don't think its worth holding back tests because of it.  I will merge if there are no objections.

miabbott:
[(3) Check that AH uses sha256 passwords as part of [a-h-t] sanity test](https://trello.com/c/bnQzGCJA/443-3-check-that-ah-uses-sha256-passwords-as-part-of-a-h-t-sanity-test)

Issue #82, opened 2017-01-24
Need to use IP(s) rather than the hostname for reboot based tests


kushaldas:
Any test that requires a reboot, for example *pkg-layering* requires IP(s) rather than a short hostname, otherwise they fail after the reboot.

```
TASK [reboot : restart hosts] **************************************************
ok: [host1]
ok: [host2]

TASK [reboot : wait for hosts to come back up (inventory_hostname)] ************
fatal: [host2 -> localhost]: FAILED! => {"changed": false, "elapsed": 120, "failed": true, "msg": "Timeout when waiting for host2:22"}
fatal: [host1 -> localhost]: FAILED! => {"changed": false, "elapsed": 120, "failed": true, "msg": "Timeout when waiting for host1:22"}
	to retry, use: --limit @/home/kdas/tmp/gotun/atomic-host-tests/tests/admin-unlock/main.retry

PLAY RECAP *********************************************************************
host1                      : ok=25   changed=15   unreachable=0    failed=1   
host2                      : ok=25   changed=15   unreachable=0    failed=1   

 [WARNING]: Consider using get_url or uri module rather than running curl
 [WARNING]: Consider using yum, dnf or zypper module rather than running rpm
```

If we use the IP directly instead of *host1* or *host2*, then the tests pass successfully.

jlebon:
Ahh heh, the `wait_for` module requires a resolvable hostname. We should probably use `{{ ansible_default_ipv4.address }}` instead of `{{ inventory_hostname }}`.

miabbott:
I'd love to get a solution for this that works for all cases.   

We've gone around on this a few times (see #9, #23, #44) and we ultimately settled on the current implementation.

Using `inventory_hostname` is what is [recommended by Ansible](https://support.ansible.com/hc/en-us/articles/201958037-Reboot-a-server-and-wait-for-it-to-come-back)

Additionally, we run these tests regularly in an OpenStack environment and haven't had any issues.

That's not to say that there aren't better ways to do this, but I feel like we have tried a number of approaches before ending up here.

jlebon:
Oh right, the OpenStack case is an issue. The nodes there have no concept of what their public IPs are. The reason this works in our OpenStack is because `inventory_hostname` is the same as the public IP.

[...]

PR #83, opened 2017-01-24
reboot: use ansible_host

jlebon:
The wait_for module requires a resolvable host. The inventory_hostname
variable however is not guaranteed to be resolvable. It may be an alias.
Use ansible_host instead, which is the actual resolvable hostname that
ansible uses to SSH, which is exactly what we want here.

Resolves: #82


miabbott:
This makes sense to me, however my testing has revealed that this change appears to break the `rpm_ostree_install` role.

It fails when the role is waiting for the host to come down.  I'm not sure why we are waiting for the host to come down in this case, but in the `reboot` role we are just waiting for the host to come back up.

I modified the role locally to remove the "waiting for host to come down" step and was able to successfully complete the role using F25AH, CAHC, and RHELAH in OpenStack.

@mike-nguyen Do we need to wait for the host to come down in that role?

mike-nguyen:
@miabbott The role does a package install with the -r flag which automatically reboots the host.  Since we can't catch any errors from the rpm-ostree install command (the command doesn't return because of the reboot), there is a check for the host to come down.  Without it, the check for the host to be up will always succeed even if there is an error.  I can modify the role to not use the -r flag, and call the reboot role as an include if everyone is okay with that.

jlebon:
Right, waiting for it to come down first is needed when using this technique. Maybe we should switch how we reboot computers to what we use in rpm-ostree:

https://github.com/projectatomic/rpm-ostree/blob/98be9cef4e8bde7f8dbaae5944c656518782fabc/tests/common/libvm.sh#L116

https://github.com/projectatomic/rpm-ostree/blob/98be9cef4e8bde7f8dbaae5944c656518782fabc/tests/common/libvm.sh#L83

I.e. we keep trying to fetch the `boot_id`, which is a unique identifier for the current boot. We stop trying when we either reach a maximum timeout (failure) or when we get a `boot_id` that's different from the previous `boot_id` (success). In practice, this works really well because we no longer have to deal with ssh port races etc...

miabbott:
I like the approach of trying to fetch the `boot_id`.  Should we scrap this PR and try to rework the `reboot` role, etc?

miabbott:
Yesterday I tried reworking the `reboot` role to use the `boot_id` method, but didn't come up with a working solution.

During that investigation, I did see the following syntax at the bottom of the examples for `wait_for` in the [official Ansible docs](https://docs.ansible.com/ansible/wait_for_module.html)

`{{ ansible_ssh_host | default(inventory_hostname) }}`

The comment on the code in the examples says:

> don't assume the inventory_hostname is resolvable

This looks like a better solution than just using `ansible_host`

jlebon:
Yeah, we can try it out. Though note that `ansible_ssh_host` has been deprecated in favour of `ansible_host`: http://docs.ansible.com/ansible/intro_inventory.html.

> `{{ ansible_ssh_host | default(inventory_hostname) }}`

This is what confuses me. This should already *be* the behaviour of `ansible_[ssh_]host`. I.e. it should already be equivalent to `inventory_hostname` if it wasn't explicitly defined in the inventory file. And that's what I've observed in my tests as well.

> This makes sense to me, however my testing has revealed that this change appears to break the `rpm_ostree_install` role.

Do you have the output of that handy by any chance? (With a couple of `-v`s tacked on, mayhaps)

miabbott:
> Do you have the output of that handy by any chance? (With a couple of -vs tacked on, mayhaps)

@jlebon https://paste.fedoraproject.org/542591/80668148/

Ran just the `rpm_ostree_install` role from the `improved-sanity-tests` against an up-to-date F25 vagrant box.

Similar result for running against a RHELAH 7.3.2 VM - https://paste.fedoraproject.org/542595/48598096/

jlebon:
Ahh I see. Because we're using `local_action`, the `ansible_host` becomes `localhost`, so it essentially waits for ssh on the localhost to go down.

I pushed a new commit that accounts for this. ⬆️ Chan you check if that works for you?

I also went over the whole codebase to make sure only `ans_reboot.yaml` is used to reboot so that it's easier to change things in the future.

miabbott:
I'm not really a fan of using `include`, but I see why it is necessary.

This worked for me on a F25AH Vagrant box and a RHELAH libvirt VM.

Thanks!

PR #85, opened 2017-02-07
selinux_verify: Use systemd to find MainPID for rpm-ostreed

cgwalters:
This way we're robust against any process renames.


jlebon:
👍

mike-nguyen:
LGTM

miabbott:
Man, guess you need to be quicker with your testing to get ahead of these folks  😆

mike-nguyen:
To be honest, I did not run the full test.  Since it was a one line change, I just verified that the command output was what we were expecting. In RHEL AH and Fedora AH, both commands still worked so I compared that output.

Issue #80, opened 2017-01-20
need tests for pulling images from v1/v2 registries using docker/atomic


miabbott:
We need to ensure we have coverage for the following matrix

`atomic pull` | `docker pull`
------------ | ---------------
v1 registry | v1 registry
v2 registry | v2 registry

miabbott:
It looks like v1 registries are dead.  And we have some additional tests to check `atomic pull` and `docker pull`.

Closed via #86

PR #86, opened 2017-02-07
roles: add role to pull/run/remove "popular" images

miabbott:
This is a really simple role that just checks we can pull, run, and
remove some of the more popular images on the Docker Hub.

Useful?  Meh.  Kind of just a checkbox because my $dayjob boss asked
me to include it.  :)

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -58,6 +58,10 @@
       tags:
         - redhat_subscription
 
+    - role: docker_pull_run_remove
+      tags:
+        - docker_pull_run_remove
+
     # Install package using package layering
     - role: rpm_ostree_install
       packages: "{{ g_pkg }}"
@@ -215,6 +219,10 @@
     - role: semanage_add_verify
       tags:
         - semanage_add_verify
+
+    - role: docker_pull_run_remove
+      tags:
+        - docker_pull_run_remove
 
     # Check layered package is still installed
     - role: rpm_ostree_install_verify
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -58,10 +58,6 @@
       tags:
         - redhat_subscription
 
-    - role: docker_pull_run_remove
-      tags:
-        - docker_pull_run_remove
-
     # Install package using package layering
     - role: rpm_ostree_install
       packages: "{{ g_pkg }}"
--- a/roles/docker_pull_run_remove/tasks/main.yml
+++ b/roles/docker_pull_run_remove/tasks/main.yml
@@ -1,0 +1,16 @@
+---
+# vim: set ft=ansible:
+#
+# popular_images is defined in roles/docker_pull_run_remove/vars/main.yml
+#
+- name: Pull the popular images from Docker Hub
+  command: "docker pull docker.io/{{ item }}"
+  with_items: "{{ popular_images }}"
+
+- name: Run the popular images
+  command: "docker run --rm docker.io/{{ item }} echo 'hello'"
+  with_items: "{{ popular_images }}"
+
+- name: Remove the popular images
+  command: "docker rmi docker.io/{{ item }}"
+  with_items: "{{ popular_images }}"
--- a/roles/docker_pull_run_remove/vars/main.yml
+++ b/roles/docker_pull_run_remove/vars/main.yml
@@ -1,0 +1,10 @@
+---
+# vim: set ft=ansible:
+#
+popular_images:
+  - alpine
+  - busybox
+  - centos
+  - debian
+  - fedora
+  - ubuntu
--- a/roles/docker_pull_run_remove/vars/main.yml
+++ b/roles/docker_pull_run_remove/vars/main.yml
@@ -4,7 +4,4 @@
 popular_images:
   - alpine
   - busybox
-  - centos
-  - debian
-  - fedora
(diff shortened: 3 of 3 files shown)


tests/improved-sanity-test/main.yml
```diff
@@ -58,6 +58,10 @@
       tags:
         - redhat_subscription
 
+    - role: docker_pull_run_remove
+      tags:
+        - docker_pull_run_remove
+
```
mike-nguyen: Since this role removes images right after pulling and running them, I don't know how much value is added by running it twice considering the amount of time it takes to pull from docker.io. WDYT?

miabbott: Yeah, I could just omit the 'removal' part and stick to pull/run.

mike-nguyen: Maybe we can  remove this role pre-upgrade and run one time after the upgrade? I think there are other tests in the suite that validate images and containers go through the upgrade successfully (i.e. cockpit).  This role validates that there are no issues pulling and running known images which we are probably only concerned about in the HEAD.

miabbott: Yeah, that seems reasonable.  I'll patch it.

miabbott:
I wonder if maybe I should rework this to be more comprehensive and cover what I wanted in #80.

Thoughts?

mike-nguyen:
The work for the additional pulls seems trivial but do/will we have access to v1 and v2 registries? If we don't have the infrastructure in place, I would probably vote against it especially  because v1 registries won't be supported going forward.  I do see value in testing atomic pull because the pulling mechanism is different. 

I've been going back and forth in my head whether we want to include these pulls in the sanity test because a single pull failure of a particular image doesn't mean that the atomic host build is bad.  Considering the time it takes to pull these images, It might be better off as a separate test.  Thoughts?

miabbott:
> The work for the additional pulls seems trivial but do/will we have access to v1 and v2 registries? If we don't have the infrastructure in place, I would probably vote against it especially because v1 registries won't be supported going forward. I do see value in testing atomic pull because the pulling mechanism is different.

The RHT registry is still v1, so that would be a big problem if users couldn't pull from that.  However, it should move to v2 eventually, so the v1/v2 test will become less valuable.

> I've been going back and forth in my head whether we want to include these pulls in the sanity test because a single pull failure of a particular image doesn't mean that the atomic host build is bad. Considering the time it takes to pull these images, It might be better off as a separate test. Thoughts?

We could trim the list to just `alpine`, `busybox`, and `ubuntu`.  They are the top 3 distro-based images on the hub (as best I can tell).  The first two are very small (less than 4MB) and the `ubuntu` one is only 130MB.

That may save us any pain related to network timeouts, etc.

mike-nguyen:
>The RHT registry is still v1, so that would be a big problem if users couldn't pull from that. However, it should move to v2 eventually, so the v1/v2 test will become less valuable.

@ddarrah said the RHT registry is now v2.  I confirmed that skopeo can inspect images from RHT.

>We could trim the list to just alpine, busybox, and ubuntu. They are the top 3 distro-based images on the hub (as best I can tell). The first two are very small (less than 4MB) and the ubuntu one is only 130MB.
That may save us any pain related to network timeouts, etc.

Sounds good to me.

miabbott:
> @ddarrah said the RHT registry is now v2. I confirmed that skopeo can inspect images from RHT.

Huh.  I did not know that, but some inspection shows it to be true.

```
$ curl -i -L -X GET https://registry.access.redhat.com/v2/
HTTP/1.1 200 OK
Date: Thu, 09 Feb 2017 15:42:33 GMT
Server: Apache
Content-Length: 2
Docker-Distribution-API-Version: registry/2.0
Connection: close
Content-Type: application/json

{}

$ curl -i -L -X GET https://registry.access.redhat.com/v1/
HTTP/1.1 404 NOT FOUND
Date: Thu, 09 Feb 2017 15:42:37 GMT
Server: Apache
Content-Length: 238
Connection: close
Content-Type: text/html; charset=UTF-8

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.</p><p>If you entered the URL manually please check your spelling and try again.</p>
```

I'll probably close out #80 as a result.

miabbott:
Well, I spoke too soon.  The RHT registry might support both v1 and v2?

```
$ curl -i -L -X GET https://registry.access.redhat.com/v1/_ping
HTTP/1.1 200 OK
Date: Thu, 09 Feb 2017 15:48:23 GMT
Server: Apache
Content-Length: 4
X-Docker-Registry-Standalone: True
X-Docker-Registry-Version: 0.6.6
X-Docker-Registry-Config: common
Connection: close
Content-Type: application/json

true
```

I'm going to guess it redirects v1 to v2 by some magic.

miabbott:
@mike-nguyen I pushed a fixup to reduce the amount of images we are pulling.

mike-nguyen:
Just going to run a quick test.

mike-nguyen:
:+1:

miabbott:
[(3) Pull popular images from Docker Hub and run them as part of [a-h-t]](https://trello.com/c/Sdg5Otee/649-3-pull-popular-images-from-docker-hub-and-run-them-as-part-of-a-h-t)

PR #89, opened 2017-02-15
iso-sanity: New test for ISO installs

cgwalters:
This turned out to be harder than I thought;
https://pagure.io/atomic-wg/issue/227

### DIFF
--- a/iso-sanity/check-iso-sanity.sh
+++ b/iso-sanity/check-iso-sanity.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+ISO=$1
+shift
+dn=$(dirname $0)
+
+tempdir=$(mktemp -d /var/tmp/isosanity.XXXXX)
+function cleanup () {
+    fusermount -u ${tempdir}/mnt || true
+    rm "${tempdir}" -rf
+}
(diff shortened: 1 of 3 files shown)


cgwalters:
Not entirely sure if this should live in this repo, since it's a fundamentally different thing than running ansible on a cloud image.  But it needs to live somewhere.

iso-sanity/check-iso-sanity.sh
```diff
@@ -0,0 +1,43 @@
...
+    fi
+    sleep 0.1
+done
+if !${success}; then
+    echo "Failed to wait for http server"
+    exit 1
+fi
+
+# --initrd-inject ${dn}/default.ks --extra-args=inst.ks=file:/default.ks
+base_args="--ram 3192 --vcpus 4 --disk size=20 --os-variant rhel7 --location http://${hostip}:${port}/"
```
jlebon: So I'm guessing we don't have to specify a ks here because anaconda automatically falls back on the default one?

cgwalters: We *do* need the above bits, but this was me trying to debug why it doesn't work even if we don't inject a kickstart.  That appears to be an anaconda behavioral change when booted from ISO vs network.

PR #90, opened 2017-02-15
Add runc tests

mike-nguyen:
This check-in add a playbook to test runc functionality.  The playbook
sets up the environment needed to run an upstream runc python test on
Atomic Hosts.

The origin plan was to write the runc python test in Ansible so we
would have a consistent testing strategy.  However, there was no straight
forward way to manipulate JSON files in Ansible to test the various
runc options.  Ideally the test would not need a modified ostree
(package layering and admin unlock).

I did run into a problem with running the upstream test through Ansible.
The test requires a tty so I used the python command line workaround.
The problem is that the workaround would always return 0 even if the
script failed so I had to check for errors in stdout.  There are checks
for the most common errors that occurred when running the test but it may
not be all inclusive.


tests/runc/main.yml
```diff
@@ -0,0 +1,135 @@
...
+      reboot: false
+      when: ansible_distribution == 'RedHat' or "'CentOS' in ansible_distribution"
+      tags:
+        - rpm_ostree_install
+
+    # Waiting on https://bodhi.fedoraproject.org/updates/FEDORA-2017-6319e78c81
+    #  to fix criu dependency
+
+    - role: rpm_ostree_install
+      packages: 'gcc python-devel redhat-rpm-config criu'
```
miabbott: The latest version of F25 (25.59) looks like it includes `criu` correctly.  I'd argue to drop this even if it means breaking the playbook for F24.

tests/runc/README.md
```diff
@@ -0,0 +1,64 @@
...
+To run the test, simply invoke as any other Ansible playbook:
+
+```
+$ ansible-playbook -i inventory tests/runc/main.yml
+```
+
+*NOTE*: You are responsible for providing a host to run the test against and the
+inventory file for that host.
+
+#### Vagrant
```
miabbott: With #91 merged, can you remove this section?

tests/runc/main.yml
```diff
@@ -0,0 +1,135 @@
...
+    - role: reboot
+      tags:
+        - reboot
+
+  post_tasks:
+    - name: Turn on user namespaces
+      shell: grubby --args="user_namespace.enable=1" --update-kernel="$(grubby --default-kernel)"
+      when: ("'CentOS' in ansible_distribution" or ansible_distribution == 'Redhat')
+
+    - include: './roles/reboot/tasks/main.yml'
```
miabbott: I'm hitting a failure here when using Ansible 2.1.4

```
TASK [include] *****************************************************************
fatal: [192.168.121.86]: FAILED! => {"failed": true, "reason": "the file_name '/home/miabbott/workspaces/projectatomic/atomic-host-tests/tests/runc/common/ans_reboot.yaml' does not exist, or is not readable"}
```


FWIW, the following warning is thrown at the beginning of the playbook.

```
[DEPRECATION WARNING]: Included file '/home/miabbott/workspaces/projectatomic/atomic-host-tests/tests/runc/./roles/reboot/tasks/../../../common/ans_reboot.yaml' not found, however since this include is not 
explicitly marked as 'static: yes', we will try and include it dynamically later. In the future, this will be an error unless 'static: no' is used on the include task. If you do not want missing includes to be 
considered dynamic, use 'static: yes' on the include or set the global ansible.cfg options to make all inclues static for tasks and/or handlers.
This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

tests/runc/main.yml
mike-nguyen: Yes!  I noticed that this morning. I was waiting for rpm dependency to get fix before I checked this in but it was taking too long.

tests/runc/main.yml
mike-nguyen: There are a lot of relative paths from the includes and roles with includes which, combined with the directory structure, makes Ansible confused.  For this particular instance, I should just include common/ans_reboot.yaml but we need to find a cleaner way of doing the includes.

miabbott:
This worked for me on CAHC, F25, and RHELAH.  Time to merge it.  Thanks @mike-nguyen for another test suite!

mike-nguyen:
Thanks for looking at it @miabbott.  I appreciate the feedback.

miabbott:
[(5) [a-h-t] Create basic test suite for runC functionality](https://trello.com/c/tTHytsuV/628-5-a-h-t-create-basic-test-suite-for-runc-functionality)

PR #91, opened 2017-02-17
docs: tidy up information regarding Vagrant

miabbott:
I happened to notice that the READMEs that we include for each of the
test suites mention running `vagrant up` to execute the playbook.
This wasn't entirely correct, since the individual tests suites don't
have a Vagrantfile in their directory.

So I removed that section from the READMEs and cleaned up the top
README to explain how to run different playbooks using the included
Vagrantfile.

Additionally, I tried to clean up the top-level README to be more
useful and correct.


mike-nguyen:
My bad!  They're all mine.

PR #92, opened 2017-02-17
Save failure details to result file

mike-nguyen:
Added a callback_plugin that makes the logs more human readable and
also saves down failure details to a file.

A majority of this code is actually from other projects.  The code
that alters the Ansible output is from openshift-ansible which is
derived from an upstream project (in the comments).  I had to override
the baseclass methods to track the failure details.  In order to
preserve the default Ansible logging, I copied the original method
code back into the override method.

All test that want to leverage the callback_plugin need a symlink to
the top callback_plugins directory otherwise the default Ansible logging
will be used.

### DIFF
--- a/callback_plugins/default.py
+++ b/callback_plugins/default.py
@@ -1,0 +1,145 @@
+'''Plugin to override the default output logic.'''
+
+# upstream: https://gist.github.com/cliffano/9868180
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+# For some reason this has to be done
+import imp
+import os
+
+ANSIBLE_PATH = imp.find_module('ansible')[1]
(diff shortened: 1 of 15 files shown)


callback_plugins/default.py
```diff
@@ -0,0 +1,145 @@
...
+                colorize(u'unreachable', t['unreachable'], None),
+                colorize(u'failed', t['failures'], None)),
+                log_only=True
+            )
+
+        self._display.display("", screen_only=True)
+
+        # Save result file in current working directory
+        try:
+            os.remove('result')
```
miabbott: I wonder if we want to make the filename more unique.  In our own Jenkins jobs, we should get a fresh workspace for each time we run the tests, but if other folks run this test multiple times from the same directory, we'll end up overwriting the file.

Perhaps prepend it with the playbook name and/or append a timestamp?

miabbott: Maybe even go the extra mile and let the filename be configurable via environment variable?

miabbott:
[(3) [CI] Report the role which failed in the email notifications for the [a-h-t] CI jobs](https://trello.com/c/SMOBZjPj/401-3-ci-report-the-role-which-failed-in-the-email-notifications-for-the-a-h-t-ci-jobs)

jlebon:
This is pretty cool. Just looking at the PR tester output, you can tell it's a drastic difference. Once suggestion might be to prefix `STDOUT` and `STDERR` lines with e.g. `> ` or maybe just a `---` header and footer to make it more easy to tell what the actual output is.

mike-nguyen:
Added an environment variable (AHT_RESULT_FILE) that, if set, will create a result file that will be blank if it passes or contain the failure task/role/output if it fails.  Also added some headers and footers to the stdout/stderr/msg.

miabbott:
The sample output from the Red Hat CI check looks great with this plugin.  I'm OK merging this assuming @jlebon has no other comments?

jlebon:
No complaints here! 👍

mike-nguyen:
Thanks!

PR #94, opened 2017-02-20
add a test for rhbz 1421867

miabbott:
Our internal testing found that trying to use `rpm-ostree status` on a
RHELAH host that was not registered via `subscription-manager` would
fail. See [RHBZ
1421867](https://bugzilla.redhat.com/show_bug.cgi?id=1421867)

To test for this, I've re-arranged when we use `rpm-ostree status` and
added a role that removes any subscriptions from a host and drops an
`unconfigured-state` field into the origin file (optionally).

### DIFF
--- a/roles/redhat_unsubscribe/tasks/main.yml
+++ b/roles/redhat_unsubscribe/tasks/main.yml
@@ -1,0 +1,42 @@
+---
+# vim: set ft=ansible:
+#
+# NOTE: This role should only be used on RHEL systems
+#
+# We get the neccessary info to insert the 'unconfigured-state' field into the origin
+# file only when we need to.  Otherwise, we just use the sub-man command to remove
+# all the subscriptions.
+#
+- block:
+  - name: Get the current commit to find the correct origin file to manipulate later...
+    command: rpm-ostree status --json
+    register: rpm_output
+
+  - name: Convert deployment information into jinja2 json
+    set_fact:
+      json: "{{ rpm_output.stdout|from_json }}"
+
+  - name: Set version and commit if deployment 0 is booted
+    set_fact:
+      booted_commit: "{{ json['deployments'][0]['checksum'] }}"
+    when: json['deployments'][0] is defined and json['deployments'][0]['booted']
+
+  - name: Set version and commit if deployment 1 is booted
+    set_fact:
+      booted_commit: "{{ json['deployments'][1]['checksum'] }}"
+    when: json['deployments'][1] is defined and json['deployments'][1]['booted']
+  when:
+    - unconfigured_state is defined
+    - unconfigured_state
+
+- name: Remove all subscriptions
+  command: subscription-manager remove --all
+
+- name: Insert 'unconfigured-state' field
+  lineinfile:
+    dest: /ostree/deploy/rhel-atomic-host/deploy/{{ booted_commit }}.0.origin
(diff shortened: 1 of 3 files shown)


miabbott:
[(2) Add test for running rpm-ostree daemon on unsubscribed system](https://trello.com/c/6o9me2xB/654-2-add-test-for-running-rpm-ostree-daemon-on-unsubscribed-system)

tests/improved-sanity-test/main.yml
```diff
@@ -210,6 +210,24 @@
...
 
+    # we remove any subscriptions after the upgrade to verify that
+    # 'rpm-ostree status' with the 'unconfigured-state' field present.
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1421867
+    - role: redhat_unsubscribe
+      unconfigured_state: true
+      when: ansible_distribution == 'RedHat'
+      tags:
+        - redhat_unsubscribe
+
```
jlebon: I would make sure to stop the daemon here before running `status`. Though from looking at the tasks above it seems like it shouldn't be running yet anyway, but it's not obvious either that we rely on that here. So I can see that assumption easily becoming false in the future.

miabbott: Do you suggest stopping the daemon as part of the role?  

The daemon will automatically start when a user runs an `rpm-ostree` command, correct?

jlebon: Yeah it will automatically start. But if it was already running, you won't be testing the path which caused that RHBZ to fail, which happened at rpm-ostree startup. Maybe make it an optional `stop_daemon_first` var to the `atomic_host_status_verify` role? It's too bad we can't mix roles and tasks, though we *are* abusing the ansible semantics here. :)

miabbott: It wouldn't be 'atomic-host-tests' if we weren't abusing Ansible.  😆

miabbott:
Pushed a new commit with support for stopping the daemon.  ⬆️

jlebon:
👍

roles/redhat_unsubscribe/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  - name: Set version and commit if deployment 1 is booted
+    set_fact:
+      booted_commit: "{{ json['deployments'][1]['checksum'] }}"
+    when: json['deployments'][1] is defined and json['deployments'][1]['booted']
+  when:
+    - unconfigured_state is defined
+    - unconfigured_state
+
+- name: Remove all subscriptions
+  command: subscription-manager remove --all
```
mike-nguyen: This step returned 70 causing the playbook to fail.

miabbott: Dang.  I actually recall hitting this yesterday and thinking I should fix it.  Will address it shortly.

miabbott:
Pushed a fixup for the `rc == 70` issue. ⬆️ 

Additionally, I improved the `redhat_unsubscribe` role further by adding the use of the `subscription-manager unregister` command.

As I understand it, a system can be subscribed to a remote service, but may not have any subscriptions attached to the system.  So using `subscription-manager remove` clears out any subscriptions locally and `subscription-manager unregister` removes the system's subscription from the remote service.

mike-nguyen:
:+1:

PR #95, opened 2017-02-28
tests: remove unmaintained tests

miabbott:
Let's face reality and remove these tests that have not been
maintained or even ran since this repo was created.  Some of the tests
are covered by the `improved-sanity-tests` and others are run manually
as part of regular releases.

And since this is `git` they are always just a revert away, in case we
decide we have made a mistake.


mike-nguyen:
Spring cleanup in full effect!  Sounds and looks good to me.  If there are no objections, I will merge today.

jlebon:
Agreed. Let's not give false impressions.

PR #96, opened 2017-03-02
atomic_scan_verify: remove upstream scan image

miabbott:
As of about 3:30pm today, there is no more
`docker.io/fedora/atomic_scan_openscap` image.  This image was part of
the old 'Fedora Dockerfiles' repo, which is being migrated to the new
Fedora Registry and OSBS.

As such, we need to revert to using the scanner from the Red Hat
registry and make a small modification to the scanner config file as a
result.


mike-nguyen:
Great work!

PR #97, opened 2017-03-03
expand SELinux tests

miabbott:
This commit expands the scope of our SELinux checks to include the
validation of the SELinux labels on multiple key processes and files.

It also checks that any files on the filesystem do not have the
default labels `default_t` or `file_t`.

Finally, it includes a simple test to validate that changes made
to an SELinux boolean will persist through the upgrade/reboot cycle.
This particular test covers [RHBZ#1381588](https://bugzilla.redhat.com/show_bug.cgi?id=1381588)

### DIFF
--- a/roles/selinux_boolean/tasks/main.yml
+++ b/roles/selinux_boolean/tasks/main.yml
@@ -1,0 +1,13 @@
+---
+# vim: set ft=ansible
+#
+- name: Fail if 'g_seboolean' is not defined
+  fail:
+    msg: "The variable 'g_seboolean' is undefined"
+  when: g_seboolean is undefined
+
+- name: Modify an SELinux Boolean
+  seboolean:
+    name: "{{ g_seboolean }}"
+    persistent: yes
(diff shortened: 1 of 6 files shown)


miabbott:
Note, the check for the default labels is going to fail across the board because of [RHBZ#1428109](https://bugzilla.redhat.com/show_bug.cgi?id=1428109)

Given that, I could break out each individual test so it could be skipped rather than lumping them all together in a single role.

miabbott:
[(3) [a-h-t] Create test(s) for verifying SELinux contexts on Atomic Host](https://trello.com/c/Zyb9FlJ2/34-3-a-h-t-create-test-s-for-verifying-selinux-contexts-on-atomic-host)

mike-nguyen:
It took me a while to figure out how the sub elements and items were being passed around in Ansible.  As abusive as you think it is, it works well!  It will simplify adding more files, procs, mountpoints, and services simple.

PR #98, opened 2017-03-03
Fix blank failure output

mike-nguyen:
There was a particular failure that caused the failure details to not
be displayed in the output of playbook test run.  Upon debugging, it
was discovered that the callback plugin was using code from Ansible 2.2.x,
which referenced a private member variable in the constructor that does
not exist in Ansible 2.1.  The code is now replaced with the default
Ansible 2.1.x code.

Having the play name  that the failed task occurred in was still needed
so I created a private variable that gets set when a new play starts.


miabbott:
This worked for me in the failure case and the success case.  Will wait on additional comments, but aiming to merge this today.

PR #102, opened 2017-03-06
Utilize ignore errors when saving failure

mike-nguyen:
The code was capturing all failures including ones that are purposely
ignored in the playbook. I changed it so it would respect ignore
errors and only save the actual failure.


miabbott:
Quick turnaround!  LGTM!

PR #104, opened 2017-03-07
Handle None case for ignore_errors

mike-nguyen:
ignore_errors is set to None when it is passed into v2_runner_on_failed
which will not capture any failures except when the calling function
does not pass an ignore_errors parameter.  I swapped it to capture the
failure in all cases except when ignore_errors is explicitly set to
true.


mike-nguyen:
I made the assumption that ignore_errors would be set to False and you know what they say when you assume.

mike-nguyen:
I tried not to dig too deep into the Ansible code but it seems like sometimes you need to.

PR #105, opened 2017-03-07
Add test for non-root package layering

mike-nguyen:
Package layering support was added for non-root packages.  There was
a negative test to make sure non-root packages could not be installed
that is now replaced with a test to make sure non root packages can
be installed.


miabbott:
[(2) Update package layering test to support non-root packages](https://trello.com/c/8EjpyrWw/673-2-update-package-layering-test-to-support-non-root-packages)

miabbott:
This actually helped find a small regression in the output of `rpm-ostree status` - projectatomic/rpm-ostree#670

The test ran fine on Fedora and looks straight-forward, so let's merge it!

PR #106, opened 2017-03-08
i-s-t: add tag for testing cloud image

miabbott:
The main outcome of this change is to introduce a tag that allows
users to run the `improved-sanity-test` playbook against a freshly
minted cloud image.  This causes portions of the playbook to be
skipped, most notably the upgraded and rollback tests.  The idea for
this addition came from the observation that Fedora runs tests via
`autocloud` against each new Atomic Host cloud image that produced
every night.

To use this new tag, a user would invoke the playbook with the `--tag
cloud_image` value:

`$ ansible-playbook -i inventory tests/improved-sanity-test/main.yml --tag cloud_image`

Currently, the order of tests/operations for this new tag should look
like this:

- compare `atomic host status` vs `rpm-ostree status`
- verify SELinux labels
- modify SELinux file contexts
- check `docker pull`, `run`, `remove`
- use pkg-layering to install a package
- use `ostree admin unlock` to install an RPM
- verify `/tmp` permissions
- verify the rpmdb is usable
- add users, modify files in `/etc/`, drop files in `/var`
- test `atomic pull`, `atomic scan`
- `docker build` and `docker run` image
- `atomic install` cockpit container
- cleanup any containers + images

As part of this effort, I re-ordered when the `osname_set_fact` role
was run and used `always_run: yes` with that role.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -38,29 +38,50 @@
 
   roles:
     # This playbook requires Ansible 2.1 and an Atomic Host
-    - { role: ansible_version_check, avc_major: "2", avc_minor: "1", tags: ['ansible_version_check'] }
+    - { role: ansible_version_check, avc_major: "2", avc_minor: "1", tags: ['ansible_version_check','cloud_image'] }
 
     - role: atomic_host_check
       tags:
         - atomic_host_check
+        - cloud_image
 
+    # Set some facts that are used in multiple roles
+    - role: osname_set_fact
+      tags:
+        - osname_set_fact
+        - cloud_image
+
+    # Verify that the output from 'atomic host status' matches 'rpm-ostree status'
     - role: atomic_host_status_verify
       tags:
         - atomic_host_status_verify
+        - cloud_image
 
+    # Verify SELinux labels are correct
     - role: selinux_verify
       tags:
         - selinux_verify
+        - cloud_image
 
+    # Verify that SELinux file contexts can be modified
     - role: semanage_add_verify
       tags:
         - semanage_add_verify
+        - cloud_image
 
     # Subscribe if the system is RHEL
     - role: redhat_subscription
       when: ansible_distribution == 'RedHat'
       tags:
         - redhat_subscription
+        - cloud_image
+
+    # Verify that 'docker pull' is successful. (Note: this is a different
+    # operation than 'atomic pull')
+    - role: docker_pull_run_remove
+      tags:
+        - docker_pull_run_remove
+        - cloud_image
 
     # Install package using package layering
     - role: rpm_ostree_install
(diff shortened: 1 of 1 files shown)


miabbott:
@dustymabe This may be of interest to you

I'd like to get #97 merged before this, because that introduces some additional roles that could get used by the `cloud_image` tag.

There is some additional re-organization and optimization that I didn't want to include in this PR, so keep your eyes out.

miabbott:
[(2) Add additional tags to 'improved-sanity-test' to allow for testing of cloud-image](https://trello.com/c/MQvSNjgm/674-2-add-additional-tags-to-improved-sanity-test-to-allow-for-testing-of-cloud-image)

tests/improved-sanity-test/main.yml
```diff
@@ -38,148 +38,180 @@
 
   roles:
     # This playbook requires Ansible 2.1 and an Atomic Host
-    - { role: ansible_version_check, avc_major: "2", avc_minor: "1", tags: ['ansible_version_check'] }
+    - { role: ansible_version_check, avc_major: "2", avc_minor: "1", tags: ['ansible_version_check','cloud_image'] }
```
mike-nguyen: When reviewing the code, I realized it was a little more difficult to read than it should be.  If its not too much trouble, maybe we can reformat everything in this style to match the consistency of the rest of the code, i.e.

- role: ansible_version_check
  avc_major: "2"
  avc_minor: "1"
  tags:
    - ansible_version_check
    - cloud_image`

tests/improved-sanity-test/main.yml
```diff
@@ -38,148 +38,180 @@
...
     # Check that the RPM database is functional
     - role: rpmdb_verify
       tags:
         - rpmdb_verify
+        - cloud_image
 
     # Add users, make changes to /etc, and add things to /var before the
     # the system is upgraded
-    - { role: user_add, ua_uid: "{{ g_uid1 }}", tags: ['user_add'] }
+    - { role: user_add, ua_uid: "{{ g_uid1 }}", tags: ['user_add','cloud_image'] }
```
mike-nguyen: Same here.

tests/improved-sanity-test/main.yml
```diff
@@ -210,7 +242,15 @@
...
         - atomic_host_check
 
-    # we remove any subscriptions after the upgrade to verify that
+    # Setup facts again. (Need to use the 'always_run' option to ensure the
+    # role runs again)
+    - role: osname_set_fact
+      tags:
+        - osname_set_fact
+        - cloud_image
+      always_run: yes
```
mike-nguyen: Good find!  We need always_run in the other tests instead of multiple tags named slightly differently.

mike-nguyen:
Ran :100:

miabbott:
Regarding the reformat of the roles, I agree that we need to update for consistency.

I wanted to do that separately so that the changes were siloed.

mike-nguyen:
@miabbott Sounds like an plan!

PR #108, opened 2017-03-09
Skip the SELinux tests temporarily

miabbott:
We know that the SELinux tests are going to fail because of the
following [RHBZ#1381588](https://bugzilla.redhat.com/show_bug.cgi?id=1381588)

So we will skip them for now.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -12,4 +12,4 @@ packages:
 
 tests:
   - ansible-playbook -vi testnode, common/ans_ah_head-1_deploy.yml
-  - ansible-playbook -vi testnode, tests/improved-sanity-test/main.yml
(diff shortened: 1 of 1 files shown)


mike-nguyen:
Good Idea!  Just have to remember to take it out when its fixed.

cgwalters:
This seems like it should have fixed https://ci.centos.org/view/Atomic/job/atomic-tree-smoketest-centos7/121/console ...

Oh, I see, it's only affecting the builds of *this* repo.  Eh, since we know they're not going to be fixed anytime soon, can we change the default to not test it?

miabbott:
> Oh, I see, it's only affecting the builds of this repo. Eh, since we know they're not going to be fixed anytime soon, can we change the default to not test it?

This makes much more sense than configuring the various places where it is being used.

Will make a PR for this today.

miabbott:
👆 #111

PR #109, opened 2017-03-13
Update tests to run on Ansible 2.2

mike-nguyen:
This check-in makes the necessary changes in the current supported
tests to run on Ansible 2.2.  The most notable change was in the
callback_plugin which is not backwards compatible with older versions
of Ansible.  Because of this, I made sure there was a check for
Ansible version 2.2 in every supported test.

There were a few roles that used the docker module that needed to be
modified to work with Ansible 2.2.  Other than that it was mostly
documentation changes.

This PR will be kept open for one week to allow time for consumers of
atomic-host-tests to update their environment since it is not backwards
compatible.


requirements.txt
```diff
@@ -1 +1 @@
-ansible==2.1.0
+ansible==2.2.0
```
miabbott: I'd change this to `ansible>2.1` that way we pick up any bug fixes or security fixes that may land in 2.2

tests/k8-cluster/main.yaml
```diff
@@ -2,47 +2,53 @@
...
+#    6. Verify that the pods are running and that the mariadb and webserver
 #       can communicate through the kubernetes services
 #
 - name: k8-cluster
   hosts: all
   become: yes
 
   roles:
+    # This playbook requires Ansible 2.2 and an Atomic Host
+    - role: ansible_version_check
```
miabbott: I'm seeing `^M` characters delimiting the lines in this file.  Can you remove them and re-submit?

http://its.ucsc.edu/unix-timeshare/tutorials/clean-ctrl-m.html

requirements.txt
mike-nguyen: @miabbott Good point about the bug fixes.  Does it make sense to use ansible>=2.2.0 to be more explicit about the requirements?

miabbott: Yeah, makes sense

miabbott:
There are a few instances of 

`Ansible version 2+ (the playbook was tested using Ansible 2.2)`

Let's just be explicit and say something like:

`Ansible version 2.2 (other versions are not supported)`

miabbott:
Oops, did not mean to close that.

miabbott:
Looking good.  Here we come Ansible 2.2!

Let's say that we'll merge this on Monday, March 20.  Sound good?

mike-nguyen:
Monday March 20th sounds good.  I updated the documentation to reflect that final issue you raised.

miabbott:
👍

miabbott:
[(3) Create a version of the [a-h-t] callback plugin that is compatible with Ansible 2.2](https://trello.com/c/6wPf1bXd/684-3-create-a-version-of-the-a-h-t-callback-plugin-that-is-compatible-with-ansible-2-2)

miabbott:
[(3) Make the [a-h-t] playbooks compatible with Ansible 2.2](https://trello.com/c/dWJuy7vL/685-3-make-the-a-h-t-playbooks-compatible-with-ansible-2-2)

miabbott:
Today's the day for the switch to using Ansible 2.2

This will be merged at approximately 18:00 UTC (14:00 EDT)

PR #110, opened 2017-03-13
semanage_add_verify: fix to run on all platforms

### LINKED ISSUE #99: old role that only runs in kubelet rpm is installed
https://github.com/projectatomic/atomic-host-tests/blob/0ad1a20/roles/semanage_add_verify/tasks/main.yml#L3

We should refocus this role or get rid of it as it only runs if kubelet rpm is installed and that is no longer an rpm in fedora or CAHC.

miabbott:
Previously we were trying to change the `fcontext` object type on a
specific directory based on if there was a `kubelet` package
installed.  This was incorrect, as I don't think any platform/stream
had that package installed.

This re-tools the role to modify an known object type to a new value
and compares the result at the end.  No uncessary checks and should
work on all streams.

Fixes #99


miabbott:
[(2) Update semanage test in [a-h-t] to use alternate directory](https://trello.com/c/PE4epFwh/678-2-update-semanage-test-in-a-h-t-to-use-alternate-directory)

roles/semanage_add_verify/tasks/main.yml
```diff
@@ -1,35 +1,26 @@
...
+- name: Retrieve original context
+  shell: "semanage fcontext -l | grep {{ test_dir|quote }}"
+  register: orig_context
 
-- name: Set fcontext for /var/lib/kubernetes
-  command: semanage fcontext -a -t {{ selinux_type }} {{ dir_match }}
-  when: have_builtin_kube
+  # NB: this may return an error message on Fedora that looks like:
+  # 'libsemanage.dbase_llist_query: could not query record value (No such file or directory).'
+  # However, the operation will still be successful.  ¯\_(ツ)_/¯
```
cgwalters: I'd guess that's because `/var/opt/quest/vas/vasd` doesn't exist in policy?  Why not use e.g. `/var/cache/ldconfig` maybe?

miabbott: The previous command where I grep the policy for that entry is successful:

```
# semanage fcontext -l | grep vasd
/var/opt/quest/vas/vasd(/.*)?                      all files          system_u:object_r:var_auth_t:s0
# echo $?
0
```

roles/semanage_add_verify/tasks/main.yml
```diff
@@ -1,35 +1,26 @@
...
+  shell: "semanage fcontext -l | grep {{ test_dir|quote }}"
+  register: mod_context
 
-- name: Fail selinux type is not in semanage output
+- name: Fail if modified context is not correct
   fail:
-    msg: "Could not find {{ selinux_type }} in output"
-  when: "{{ selinux_type }} not in output.stdout and output is defined"
+    msg: "The SELinux context does not appear to be changed!"
+  when: test_context not in mod_context.stdout or orig_context.stdout == mod_context.stdout
```
miabbott: Dang.  That is right.  Might have to split out the verification part.

miabbott:
Pushed a new commit that significantly alters the roles. ⬆️

mike-nguyen:
LGTM.  Tested on F25 and RHEL.

PR #111, opened 2017-03-13
Re-enable SELinux tests

miabbott:
Rather than configure all the executors of the tests to skip an entire
role, we'll just comment out the failing part of the role until the
necessary bug ([RHBZ #1381588](https://bugzilla.redhat.com/show_bug.cgi?id=1381588)) is fixed.

### DIFF
--- a/roles/selinux_verify/tasks/main.yml
+++ b/roles/selinux_verify/tasks/main.yml
@@ -78,13 +78,11 @@
   register: find_default_t
   with_items: "{{ mountpoints }}"
 
-# Until RHBZ #1381588 is fixed, we are commenting out this test rather than
-# excluding the whole role from being run by various executors.
-#- name: Fail if a file/dir is found with 'default_t' label
-#  fail:
-#    msg: "The file {{ item.item }} had an SELinux label of 'default_t'"
-#  when: "{{ item.stdout | length }} != 0"
-#  with_items: "{{ find_default_t.results }}"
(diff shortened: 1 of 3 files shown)


cgwalters:
Seems sane to me.

roles/selinux_verify/tasks/main.yml
```diff
@@ -78,11 +78,13 @@
...
-    msg: "The file {{ item.item }} had an SELinux label of 'default_t'"
-  when: "{{ item.stdout | length }} != 0"
-  with_items: "{{ find_default_t.results }}"
+# Until RHBZ #1381588 is fixed, we are commenting out this test rather than
+# excluding the whole role from being run by various executors.
+#- name: Fail if a file/dir is found with 'default_t' label
+#  fail:
+#    msg: "The file {{ item.item }} had an SELinux label of 'default_t'"
+#  when: "{{ item.stdout | length }} != 0"
+#  with_items: "{{ find_default_t.results }}"
```
mike-nguyen: Do we know if the other mountpoints have the same failure?  Would it be better if we just commented out /sysroot from the vars file so we retain the checks for the other mountpoints?

miabbott: Hmm...`/sysroot` is the majority of the filesystem so I don't think it would matter too much if we used the current approach.

But I guess checking some of the filesystem rather than none of it would be preferable.

mike-nguyen:
Looks good!

PR #113, opened 2017-03-16
Change to more complex rpm-ostree install package

mike-nguyen:
A new feature was added to allow non-root packages to be layered
using rpm-ostree install.  Instead of using a simple package like
wget, the httpd package  tests a broader code path because it has
multiple dependencies, creates an apache user, and is a non root
package.

The default settings for httpd conflicted with the docker container
tests in improved sanity tests so it was changed to port 8080 in the
httpd_start role.  The httpd service is validated by retrieving the
default apache page.

### DIFF
--- a/roles/httpd_start/tasks/main.yml
+++ b/roles/httpd_start/tasks/main.yml
@@ -1,0 +1,16 @@
+---
+# vim: set ft=ansible:
+#
+# configures httpd port and starts service
+#
+- name: Change port in http conf file
+  replace:
+    dest: /etc/httpd/conf/httpd.conf
+    regexp: '^Listen 80$'
+    replace: 'Listen {{ port }}'
+  when: port is defined
+
+- name: Start httpd
+  service:
+    name: httpd
+    state: started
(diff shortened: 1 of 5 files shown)


mike-nguyen:
Note:  The improved sanity test will not pass unless the starting version contains rpm-ostree support for layering non-root packages.

cgwalters:
It should be possible to use the new [YAML --version](https://github.com/projectatomic/rpm-ostree/pull/665) to conditionalize (if `--version` doesn't output YAML it's too old).

mike-nguyen:
@cgwalters That'll be really useful when all the streams get the newer versions

miabbott:
LGTM.  I'll wait on merging based on the comment here - https://trello.com/c/LjwJQYOV/663-5-expand-rpm-ostree-package-layering-tests-in-a-h-t-to-cover-more-scenarios#action-58cb1aba8278356e5f83d321

cgwalters:
It should be pretty easy to check whether or not `rpm-ostree --version` is YAML; one approach is to check whether the first line of output has a `:`.  Then treat the non-YAML case as "older than v2017.4".

mike-nguyen:
bot, retest this please

miabbott:
[(5) Expand rpm-ostree package layering tests in [a-h-t] to cover more scenarios](https://trello.com/c/LjwJQYOV/663-5-expand-rpm-ostree-package-layering-tests-in-a-h-t-to-cover-more-scenarios)

miabbott:
LGMT! 👍

PR #114, opened 2017-03-17
new test for validating base images with Atomic Host

miabbott:
(That's a bad summary of the changes, but I couldn't do better in less
than 70 characters.)

This test intends to validate that any changes to the Atomic Host
platform will not break the usage of any 'reasonably popular' base
images.  This is accomplished by building an `httpd` container on top
of each base image, running the resulting image, and verifying the
content returned from the container is correct.

I chose 'popular' base images from Docker Hub based on the number of
pulls.  I also threw in the Fedora images from the new Fedora
registry and the RHEL images from the Red Hat registry.


tests/docker-build-httpd/main.yml
```diff
@@ -0,0 +1,81 @@
...
+  hosts: all
+  become: yes
+
+  # This is a hack.  Using the 'redhat_subscription' role as an 'include' was
+  # throwing an error about the 'subscription_file' variable not being
+  # defined, despite there being 'roles/redhat_subscription/vars/main.yaml'
+  # with the variable defined there.  This kind of short-circuits the role
+  # by providing the value at the playbook level.
+  vars:
+    subscription_file: 'roles/redhat_subscription/files/subscription_data.csv'
```
mike-nguyen: Does it make sense to have multiple plays in this scenario?  Maybe a 'setup' play that will do the subscription with post_task to manage the variables pull the images and then have a separate play that runs the test.  I'm not sure if this makes sense in this situation because way you have it now is much cleaner.

miabbott:
Force pushed changes to separate the sections of the playbook into setup, tests, teardown. ⬆️

miabbott:
[(3) Migrate 'docker build' playbook to upstream [a-h-t] repo](https://trello.com/c/Ny3U2tIG/682-3-migrate-docker-build-playbook-to-upstream-a-h-t-repo)

mike-nguyen:
:100:

PR #118, opened 2017-03-24
Small fixes for Ansible 2.2 support

miabbott:
When running playbooks that use the `always_run` option¸ Ansible 2.2
will print a deprecation warning that states that users should switch
to `check_mode: no` instead.

Additionally, another warning is printed about using `port` as a
parameter to role.  The offending role was `httpd_start` and it was
updated to use a different parameter name.


mike-nguyen:
LGTM.  Merging!

PR #119, opened 2017-03-27
Move 'atomic scan' tests ahead of 'ostree admin unlock'

miabbott:
We have been seeing multiple failures during the `atomic scan` tests
because of an incompatibility between overlay mounts and SELinux.
This won't be fully fixed until RHEL 7.4 ships, so we'll just work
around it for now.


mike-nguyen:
Pretty straight forward.  LGTM.

PR #117, opened 2017-03-22
Cleanup unused files and directories

miabbott:
The original creation of this repo did not have a clear direction or
set of guidelines on how we wanted to structure the playbooks.  As
such, there were a number of extra directories and files that were
created as we found our way.

We are in a better place now and have a better sense of how we want
the repo to look.  This is the first of a few commits that will
cleanup the repo and hopefully make it less of a mess.

PR #122, opened 2017-03-28
Consistently use .yml as the suffix for YAML files

miabbott:
The repo had files with both `.yaml` and `.yml` suffixes, which could
lead confusion when trying to include files as part of a playbook.
This is part of the OpenShift Ansible style guide, which we try to
stick to as much as possible.

As part of the rename, all references or `include:` statements were
updated to point to the correct files.

There may have been a few typos corrected or extra white spaces
removed along the way.

This should be merged after #117.


mike-nguyen:
:eyes: :+1: :two:  :nerd_face:

miabbott:
[(3) Clean up [a-h-t] repo](https://trello.com/c/Vur1nNpV/692-3-clean-up-a-h-t-repo)

PR #123, opened 2017-03-30
Add status table to repo landing page

mike-nguyen:
This commit adds a status table of the latest improved sanity test
runs for the streams we test.  It uses shields.io badges containing
the version and result of the test run.  The badges are not directly
from shields.io, instead badges and logs are saved in a s3 bucket.


mike-nguyen:
See how it looks here: https://github.com/mike-nguyen/atomic-host-tests/tree/dashboard

miabbott:
This is awesome!

jlebon:
Indeed, very cool!

PR #124, opened 2017-03-30
Handle unreachable scenario in callback plugin

mike-nguyen:
Captures the unreachable message in the result array when there is
a connection error while running a playbook.


miabbott:
I'm no expert in callback plugins, but this seems OK to me.  :shipit:

PR #129, opened 2017-04-07
Cleanup some of the reboot usage

miabbott:
I was debugging what turned out to be a user error and saw that we
actually called the `reboot` role twice in the `pre_upgrade` stage.

While I was in there, I thought it would make sense to diverge from
the typical tag naming guidelines and give each `reboot` role a unique
tag.


mike-nguyen:
LGTM!

PR #132, opened 2017-04-11
Fix k8-cluster test for Fedora

mike-nguyen:
The k8-cluster test was not run against Fedora for quite some time.
The test was failing and after some debugging it turned out that the
python package was removed from the Fedora latest container causing
cgi-bin/action in the web server to not be able to connect with the
mariadb instance.  Python3 is still available in the Fedora container.

There were two solutions: convert the action script to python3 or
install python in the Dockerfile.  I chose the latter because the action
script was used in documentation and wanted to keep it consistent.

The Dockerfile and content was compressed in a tgz file which makes it
difficult to make changes to the content.  The tgz was extracted to
facilitate future changes to the content.  The appropriate changes were
made to the role to support the new file structure.


mike-nguyen:
https://trello.com/c/vd40kSOe/710-3-fix-a-h-t-k8-cluster-test

mike-nguyen:
[(3) Fix [a-h-t] k8-cluster test](https://trello.com/c/vd40kSOe/710-3-fix-a-h-t-k8-cluster-test)

miabbott:
The `FROM` lines in the Dockerfiles should be fully-qualified.  Additionally, I'd pin to a specific major version so we don't get bit by any changes when `:latest` points to a newer version of something with newer dependencies.

I'd say the following registry sources would be best:

```
registry.fedoraproject.org/fedora:25
registry.access.redhat.com/rhel:7.3
docker.io/centos:7
```

mike-nguyen:
I made the changes in the Dockerfile and retested on RHEL AH, CentOS, and Fedora 25.  Thanks for the recommendations of the registry sources!

miabbott:
This looks good.  One nit is that explicitly using `/tmp` as a destination directory is predictable.  It is usually better to use `mktemp -d` to create a directory and use that instead.

However, I think any risk for using `/tmp` for tests is pretty low.

PR #135, opened 2017-04-19
Add additional tests for package layering

mike-nguyen:
The functionality for package layering has changed slightly so the
test need to be updated.  Two tests were added to verify the new
fucntionality.  The first test verifies that packages that are
being layered is added to a list of requested packages even if it is
already in the deployment.  The second test is that rpm-ostree should
error out when attempting to layer a package that already exists.

There was also a change in the rpm-ostree status output.  When there
are no packages are layered, there will be no "packages" listed in
the rpm-ostree status output. This caused an error when setting facts
in  rpm_ostree_uninstall_verify role.  To get around this, I only
set the fact to {} if it is undefined or to the list of packages if
it is defined.

### DIFF
--- a/roles/rpm_ostree_uninstall_verify/tasks/main.yml
+++ b/roles/rpm_ostree_uninstall_verify/tasks/main.yml
@@ -28,6 +28,9 @@
   set_fact:
     installed_json: "{{ installed.stdout | from_json }}"
 
+# This is a hack for packages being displayed in rpm-ostree v2017.3 that should
+# be fixed in v2017.4.  Once all streams are at v2017.4 we can fix this.  See
+# https://github.com/projectatomic/rpm-ostree/pull/670
 - name: Set packages when deployment 0 is booted
   set_fact:
     installed_pkgs: "{{ installed_json['deployments'][0]['packages'] if installed_json['deployments'][0]['packages'] is defined else {} }}"
--- a/roles/rpm_ostree_uninstall_verify/tasks/main.yml
+++ b/roles/rpm_ostree_uninstall_verify/tasks/main.yml
@@ -30,13 +30,13 @@
 
 - name: Set packages when deployment 0 is booted
   set_fact:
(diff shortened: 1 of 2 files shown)


roles/rpm_ostree_uninstall_verify/tasks/main.yml
```diff
@@ -30,13 +30,13 @@
 
 - name: Set packages when deployment 0 is booted
   set_fact:
-    installed_pkgs: "{{ installed_json['deployments'][0]['packages'] }}"
-  when: installed_json['deployments'][0]['packages'] is defined and installed_json['deployments'][0]['booted']
+    installed_pkgs: "{{ installed_json['deployments'][0]['packages'] if installed_json['deployments'][0]['packages'] is defined else {} }}"
```
jlebon: Though note this should be fixed in v2017.4 (see https://github.com/projectatomic/rpm-ostree/pull/670). We did resping v2017.3 in Fedora at least to include that patch, though it's probably still present in some other streams. Maybe a small comment that we should be able to drop this hack once all the streams have v2017.4?

tests/pkg-layering/main.yml
```diff
@@ -334,13 +334,31 @@
...
+      command: rpm-ostree status --json
+      register: req
+
+    - name: Convert rpm-ostree status output to jinja2 json
+      set_fact:
+        req_json: "{{ req.stdout | from_json }}"
+
+    - name: Set packages when deployment 0 is booted
+      set_fact:
+        req_pkgs: "{{ req_json['deployments'][0]['requested-packages'] if req_json['deployments'][0]['requested-packages'] is defined else {} }}"
```
jlebon: Similarly for `requested-packages` entries.

mike-nguyen:
Added some comments about the hack until all streams are at v2017.4.  I will fix the become issue in a separate PR.

miabbott:
Good fixes!  Let's not lose track of the non-privileged user change that needs to be made.

PR #136, opened 2017-04-19
Verify rpm-ostree status as non-root

mike-nguyen:
Adds a check for running rpm-ostree status as a non-root user pre
and post upgrade.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -56,6 +56,13 @@
         - osname_set_fact
         - cloud_image
 
+    # Verify that rpm-ostree can be run as non-root or sudo
+    - role: rpm_ostree_status
+      become: no
+      tags:
+        - rpm_ostree_status
+        - cloud_image
+
     # Verify that the output from 'atomic host status' matches 'rpm-ostree status'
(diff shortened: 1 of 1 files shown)


mike-nguyen:
https://trello.com/c/LoGB1stD/723-2-add-test-to-a-h-t-to-verify-non-root-users-can-use-rpm-ostree-status

miabbott:
Easy fix.  Closes #136

jlebon:
This depends on the ansible user being used though, right? If we're already root, then `become: no` will still run as root. Maybe we should instead use `become_user` with e.g. `nobody` or something?

jlebon:
rpm-ostree uses `bin`: https://github.com/projectatomic/rpm-ostree/blob/master/tests/vmcheck/test-layering-basic.sh#L52

miabbott:
⬆️ This guy has some good ideas.

mike-nguyen:
^^ seriously, I had not thought of that

mike-nguyen:
I'll test some stuff out and submit a new patch

cgwalters:
This sort of duplication is exactly why I filed https://github.com/projectatomic/atomic-host-tests/issues/74

PR #137, opened 2017-04-19
Assume user could be root when checking privileges

mike-nguyen:
In #136 a test was added to check that a non-root user could run rpm-ostree
status.  There was an assumption made that the user running the test was
not a root user.  It is bad to assume.  This check-in will now run rpm-ostree
status as bin using runuser.

To facilitate additional privilege checks, two new roles were added that will
verify that commands can/cannot be run by a specific user.


tests/improved-sanity-test/main.yml
```diff
@@ -324,10 +325,11 @@
...
     # Verify that rpm-ostree can be run as non-root or sudo
-    - role: rpm_ostree_status
-      become: no
+    - role: command_privileged_verify
+      cmd: rpm-ostree status
+      user: bin
       tags:
-        - rpm_ostree_status
+        - command_privileged_verify
         - cloud_image
```
miabbott: On line 333, I don't think we need the second `cloud_image` tag.  Since no upgrade is taking place, we would just be running the same test twice in the cloud image case.  (This isn't technically part of this review, but I just noticed it now).

miabbott:
TIL about `runuser`

LGTM

mike-nguyen:
I removed the second cloud_image tag.  TIL about `runuser` from @jlebon

roles/command_unprivileged_verify/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
```
jlebon: Seems like we can combine the two roles and just take another parameter, e.g. `expect_failure` and have something like:

```yaml
failed_when: (ruo.rc == 0 and expect_failure) or (ruo.rc != 0 and not expect_failure)
```

?

mike-nguyen: @jlebon Good idea.  I'll give it a whirl.

miabbott:
Nice and concise!  @jlebon is full of good ideas!  💯

PR #138, opened 2017-04-20
roles: modify 'selinux_verify' to be distro aware

### LINKED ISSUE #133: Special case Fedora 26 selinux context for /usr/bin/docker-storage-setup
Container-storage-setup is used in F26 which makes /usr/bin/docker-storage-setup a symlink with the selinux context of system_u:object_r:bin_t:s0.  Dan Walsh said "docker-storage-setup is a symbolic link to container-storage-setup So bin_t is fine."

The selinux_verify role currently checks for container_runtime_exec_t so there needs to be a special case for Fedora 26.

miabbott:
This changes how the `selinux_verify` role will check the SELinux file
and process labels.  The main improvement is the ability to create
distro-specific checks of the labels.

Instead of one master file that contains all the files and processes
we want to check, there is now two locations where this information
is defined.  There is a 'common' file that contains the files/processes
that can be found on all supported streams.  And there are distro-specific
(per major-release) files that define files/processes that which don't
land in the 'common' set.  This brings in great flexibility for
supporting new versions of streams, as well as the ability to expand
the files/processes we want to check.

This change also does away with the complexity of the
`with_subelements` method that was used before and uses `with_items`
which is easier to understand.  The drawback to this approach is some
duplication of key/values, but I think the overall gain is worth it.

Closes #133


miabbott:
[(3) Fix SELinux context test in [a-h-t] to properly support F26](https://trello.com/c/q9BtvI8O/724-3-fix-selinux-context-test-in-a-h-t-to-properly-support-f26)

mike-nguyen:
The vars with with_first_found is a pretty neat trick.  :+1:

PR #140, opened 2017-04-26
redhat_subscription: rework handing of origin files

miabbott:
This changes how we manipulate the origin files by using `ostree admin
--print-current-dir` to derive the origin file.  This gains us more flexibility because the origin file does not always live at `<commit>.0.origin`.


miabbott:
Added another commit that modified the `redhat_unsubscribe` role to be more streamlined, as well.

mike-nguyen:
LGTM

PR #141, opened 2017-04-27
Check for rootfs dir in remote sys container

mike-nguyen:
Update the test to reflect the symlink of the rootfs dir when using
--rootfs.  See https://github.com/projectatomic/atomic/pull/922


mike-nguyen:
Note: The run_state_dir test will still fail until the atomic host picks up: https://github.com/projectatomic/atomic/pull/962

use --skip-tags run_state_dir

miabbott:
Took more work to write the commit message than the fix itself!

Works for me.

PR #142, opened 2017-04-27
Cleanup after test completion

miabbott:
If we want to run the tests in succession, we should cleanup what the
tests have left around.  This typically includes docker
images/containers, kubernetes artifacts, and layered packages.

This change attempts to cleanup after most* tests and return the host
under test to its original state.  This includes two new roles that
assist in the cleanup of the host.

* The improved-sanity-test and system-containers test were not
addressed in this change, but should be addressed in the future.

### DIFF
--- a/tests/runc/main.yml
+++ b/tests/runc/main.yml
@@ -137,3 +137,31 @@
         'exit status 1' in result.stdout or
         'error' in result.stdout or
         'Traceback' in result.stdout
+
+- name: runc - Cleanup
+  hosts: all
+  become: yes
+
+  tags:
+    - cleanup
+
+  vars_files:
+    - vars.yml
+
+  roles:
+    - role: rpm_ostree_rollback
+      tags:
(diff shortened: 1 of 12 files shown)


miabbott:
Unfortunately, if the playbook fails during execution, these cleanup routines don't run.

miabbott:
Found some extra cleanup that needs to happen. ⬆️

miabbott:
OK, I addressed the system containers test now.

tests/runc/main.yml
```diff
@@ -137,3 +137,31 @@
...
+
+  roles:
+    - role: rpm_ostree_rollback
+      tags:
+        - rpm_ostree_rollback
+
+    - role: reboot
+      tags:
+        - reboot
+
```
miabbott: Good catch!  Fixed up ⬇️

mike-nguyen:
LGTM

PR #143, opened 2017-05-01
Start using images from the Fedora registry

miabbott:
We should start using system containers from the Fedora registry for
non-RHEL systems.


cgwalters:
LGTM!

mike-nguyen:
Looks good!

PR #144, opened 2017-05-03
Collect journal on failure

mike-nguyen:
This commit will collect the journal for the improved sanity test and
save it locally.  It also enables running other tasks on failure.

This is a giant workaround since Ansible does not have direct support to
run post execution tasks on failure..  Ansible's only mechanism for running
tasks after a play completes is a handler.  The problem with a handler
is that it will not run on failure without the force_handlers set to true.
If force_handlers is set to true it will run after every block of
pre_tasks, post_task, roles, and plays (not desirable).  The way to
workaround that is to conditionalize the tasks to only run when there
is a failure; however, Ansible does not have a direct variable that holds
the pass/fail status of a playbook.

In order to get the fail status, I leveraged the v2_runner_on_failed
method in the callback plugin that is called when Ansible fails.
An environment variable with the fail status of the result. Now the
handler tasks can be conditionalized to run on failure based on the
environment variable.

The handler can notify tasks individually by name or tasks can listen for
a notification.  Since multiple tasks need to be run, I opted to have
all tasks that need to be run together to listen for the same notification.
Notified tasks are run in the order they coded and not by the time they
are notified.

### DIFF
--- a/callback_plugins/default.py
+++ b/callback_plugins/default.py
@@ -84,9 +84,13 @@ class CallbackModule(DEFAULT_MODULE.CallbackModule):  # pylint: disable=too-few-
             self._display.display("fatal: [%s]: UNREACHABLE! => %s" % (result._host.get_name(), self._dump_results(result._result)), color=C.COLOR_UNREACHABLE)
 
     def v2_runner_on_failed(self,result, ignore_errors=False):
-        # Save last failure and set environment variable
         if ignore_errors is not True:
+            # Sets environment variable for test failures for use in playboks.
+            # Handlers tasks can conditionalize themselves using this variable
+            # to run only on failure.
             os.environ["AHT_FAILURE"] = "1"
+
+            # Save last failure
             self.failed_task = result
 
         if self._play.strategy == 'free' and self._last_task_banner != result._task._uuid:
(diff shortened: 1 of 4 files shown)


miabbott:
Great work here!  Implementing the handler in each set of tests is a bit complex, but I think the flexibility that is built in to the generic handler is excellent.  I'm going to test this today and hopefully get this merged soon.

callback_plugins/default.py
```diff
@@ -84,8 +84,9 @@ def v2_runner_on_unreachable(self, result):
             self._display.display("fatal: [%s]: UNREACHABLE! => %s" % (result._host.get_name(), self._dump_results(result._result)), color=C.COLOR_UNREACHABLE)
 
     def v2_runner_on_failed(self,result, ignore_errors=False):
-        '''Save last failure'''
+        # Save last failure and set environment variable
```
miabbott: I would expand this comment a bit more to explain how this variable is used by the handlers.

miabbott:
This didn't impact the `improved_sanity_test` in the successful case (which is good) and I was able to hack up a quick test playbook that test the failure case.

However, I wasn't able to get it working when including the role in a `pre_tasks` or `post_tasks` section.  Perhaps some explicit examples would be useful.

mike-nguyen:
If used as a include instead of a role, you have to add the handlers to the end of the play.  Using as a role automatically pulls in the handlers in the role directory.

```
handlers:
  - include: path/to/your/handlers.yml
```

miabbott:
Gonna merge this, but we could use some more docs on how to use it in different scenarios.

miabbott:
[(5) [CI] Capture the journal from a system when [a-h-t] test fails](https://trello.com/c/LjfZiLEv/402-5-ci-capture-the-journal-from-a-system-when-a-h-t-test-fails)

PR #145, opened 2017-05-05
roles: rework atomic_installation_verify

miabbott:
While testing an `atomic` update manually, I hit an issue
(projectatomic/atomic#995) that I thought our tests should have
caught.  And when I examined this role as a response, I realized there
were some improvements to be made.

- Switch to using the cockpit image from the Fedora registry
- Stop checking the cockpit version (turns out the version check never
worked)
- Run the same commands twice; one specifying the `:latest` tag and
once without a tag
- Stop the container before uninstalling


mike-nguyen:
LGTM

PR #146, opened 2017-05-08
roles: fix selinux_verify for F24

miabbott:
I got tired of seeing the Fedora 24 test failing on SELinux checks, so
I created the necessary config for it.


mike-nguyen:
LGTM

PR #151, opened 2017-05-18
Add test + roles to verify docker RPMs can be built

cevich:
Support two test-cases for docker-current and docker-latest
on RHEL or Fedora:

1) Can the current .spec file + upstream sources, produce
   installable and minimaly functional docker RPMs.

-or-

2) Can #1 be achieved with an alternate .spec file
   and/or alternate upstream sourc(es) (e.g. a commit
   to a branch by a PR)

### DIFF
--- a/tests/docker-rpm/README.md
+++ b/tests/docker-rpm/README.md
@@ -1,0 +1,47 @@
+This playbook builds the docker-latest RPMs, installs them,
+then verifies the basic docker functions.
+
+Core Functionality
+  - Building RPMs from spec + source
+  - Building RPMs from spec + alternate source (e.g. a PR)
+  - Install the built RPMs
+  - docker build
+  - docker images
+  - docker ps
+  - docker pull
+  - docker rm
+  - docker rmi
+  - docker run
+  - docker start
+  - docker stop
+
+### Prerequisites
+  - Ansible version 2.2 (other versions are not supported)
+
+  - A RHEL or Fedora system
+
+  - Configure subscription data
+
+    If running against a RHEL, you should provide subscription
+    data that can be used by `subscription-manager`.  See
+    [roles/redhat_subscription/tasks/main.yml](roles/redhat_subscription/tasks/main.yml)
+    for additional details.
+
+  - Some other stuff
+
+  - Configure the required variables to your liking in [tests/docker-rpm/vars.yml](vars.yml).
+
+### Running the Playbook
+
+To run the test, simply invoke as any other Ansible playbook:
+
+```
+$ ansible-playbook -i inventory tests/docker-rpm/main.yml -e subscription_file="/path/to/subscription_data.csv"  -e @/path/to/vars.yml
+```
+
+By default, this test will test docker-latest.  If you would like to test the
+out-of-box release of docker, set `g_docker_latest` to false
+in [tests/docker-rpm/vars.yml](vars.yml).
+
+*NOTE*: You are responsible for providing a host to run the test against and the
+inventory file for that host.
--- a/tests/docker-rpm/builder.sh
+++ b/tests/docker-rpm/builder.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+. env.sh
+
+case "$PACKAGE" in
+    docker)
+        . pkgs/$PACKAGE.sh
+        ;;
+    docker-latest)
+        . pkgs/$PACKAGE.sh
+        ;;
+    *)
+        . pkgs/default.sh
+        ;;
+esac
+
+update_sources_and_spec
+
+. common.sh
+cleanup_stale
+fetch_and_build
+
+if [ $BUILDTYPE == "tagged" ]; then
+    commit_to_dist_git
+fi
--- a/tests/docker-rpm/callback_plugins
+++ b/tests/docker-rpm/callback_plugins
@@ -1,0 +1,1 @@
+../../callback_plugins
--- a/tests/docker-rpm/common.sh
+++ b/tests/docker-rpm/common.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+
+. env.sh
+
+# delete stale packages, tarballs and build dirs
+cleanup_stale ()
+{
+    pushd $PKG_DIR/$PACKAGE
+    git clean -dfx
+    popd
+}
+
+# update spec changelog and release value
+bump_spec ()
+{
+    export CURRENT_VERSION=$(cat $PACKAGE.spec | grep -m 1 "Version:" | sed -e "s/Version: //")
+    if [ $CURRENT_VERSION == $VERSION ]; then
+        rpmdev-bumpspec -c "$(cat /tmp/$PACKAGE.changelog)" $PACKAGE.spec
+    else
+        rpmdev-bumpspec -n $VERSION -c "$(cat /tmp/$PACKAGE.changelog)" $PACKAGE.spec
+        sed -i "s/Release: 1\%{?dist}/Release: 1.git\%{shortcommit0}\%{?dist}/" $PACKAGE.spec
+        sed -i "s/$VERSION-1/$VERSION-1.git$SHORTCOMMIT/1" $PACKAGE.spec
+    fi
+}
+
+# rpmbuild
+fetch_and_build ()
+{
+    pushd $PKG_DIR/$PACKAGE
+    git checkout $DIST_GIT_TAG
+    bump_spec
+    spectool -g $PACKAGE.spec
+    sudo $BUILDDEP $PACKAGE.spec -y
+    rpmbuild -ba $PACKAGE.spec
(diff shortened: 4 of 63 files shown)


cevich:
**DNR:** Just a stub for now.

miabbott:
Do we want to get into the business of building the artifacts we are testing?

Up to this point, our team has been operating that under the assumption that we would test artifacts from some other source, maintaining a separation of building and testing.  We want the tests to be plugged into a system that reacts to receiving a new artifact.  While it is possible to combine the building of artifacts and testing of artifacts, that seems like we would open up ourselves to owning more work that we want to.

I know this won't affect our current way of doing things, but I am cautious about entering into this landscape.

cevich:
>While it is possible to combine the building of artifacts and testing of artifacts, that seems like we would open up ourselves to owning more work that we want to.

A totally valid concern.  My intention is to involve both RCM (heavily) and Engineering (lesser) in both writing and maintaining this.  It absolutely won't work w/o their buy-in, coding assistance, and ongoing support.

Testing wise, we still need to test the "official" builds from the build systems.  Having build logic encoded as modular/reusable roles allows providing rapid feedback (minutes instead of hours/days) to engineering.  It also opens up the possibility of re-using the build-related roles, in the build systems and other existing use-cases.  One common source and collection of processes instead of 50 disparate sources and TONs of duplicated code.

If we can keep most of the new roles somewhat generalized, it will also better support long-term use beyond just docker packages.  Enabling building + testing of other atomic-host / container related packages on multiple platforms and under multiple external projects.

Contingency Plan:  If I/we fail to get wider ownership (RCM especially) PR is closed and I/we figure something else out.  This is also why I haven't written any actual code yet 😀 

Does that address your concerns?

miabbott:
> Testing wise, we still need to test the "official" builds from the build systems. Having build logic encoded as modular/reusable roles allows providing rapid feedback (minutes instead of hours/days) to engineering. It also opens up the possibility of re-using the build-related roles, in the build systems and other existing use-cases. One common source and collection of processes instead of 50 disparate sources and TONs of duplicated code.

I would turn that on its head and ask why aren't the folks building the software just pulling in the tests from here and running them as part of the post-build process?

Perhaps this is just a personal hangup of mine, but having the necessary build logic/roles/software living in a repo that has (so far) been only for tests feels icky.  Maybe I need to let the kids on my lawn and stop yelling at them to get off.  😄

I don't want to be a roadblock to progress, so I'm not going to outright say NO to this proposal.  But we should keep tabs on this and make sure we aren't heading in a direction we don't want to go towards.

Thanks for the civilized discourse! 👍

cevich:
> I would turn that on its head and ask why aren't the folks building the software just pulling in the tests from here and running them as part of the post-build process?

Smart question.  Two reasons:

1) Building is far less automated than we'd like (esp. re: upstream source PRs).
2) The current automated parts of building [docker] needs a new home (somewhere).

> icky

IKR, and "all code sucks" and most of it is quite messy also.  We already have the "how to build" details stored and parallel-maintained in five separate places:

1. The Fedora dist-git (a.k.a. ``.spec`` file) repository
1. The RHEL dist-git repo.
1. The CentOS dist-git repo.
1. The docker repository itself (i.e. make)
1. Within the (current) "automated" build script (that needs a new home)

Reducing that by one is a 20% improvement 😀 

> fluffy, undisturbed grass

Building and testing very frequently go together.  Normally that's limited to unittesting, but the conundrum with docker is, building doesn't produce just one thing, and those things have have complex interactions with the rest of the environment (kernel, networking, storage, namespaces, oh my!).

At the same time, I do see your point, and we'll know we've done very wrong if the outcome is needing to rename this repo "KitchenSink".

> Thanks for the civilized discourse!

Absolutely, and it needs to go a bit further too.  I think we need to have conditions for working on this as:

1. All work may be discarded / thrown away (no crying!)
1. Merging must not leave dirt-bike ruts in Micah's lawn (all blades must remain green, sharp, and skyward-pointing)
1. It might simply not work and/or require too much duct-tape (nobody wants that)
1. Must have buy in for implementation and long-term maintenance across multiple groups/people  (not just a QE thang).

Anything else?

miabbott:
> Anything else?

I think we have a good idea of ground rules and expectations between the two of us, but there are more folks involved here.  I would like to hear from them, if they so choose.

cevich:
I can't use the github widget to link directly to this PR from Trello (since I don't have write permission and Micah's not on our board).  In any case, here's the trello card backing this:

https://trello.com/c/5fpnCo9O

miabbott:
[(CN) [FIAT] Move existing docker PR testing scripts to public location with wide access](https://trello.com/c/5fpnCo9O/500-cn-fiat-move-existing-docker-pr-testing-scripts-to-public-location-with-wide-access)

miabbott:
@cevich Got it all linked up for you ⬆️

cgwalters:
Generating RPMs from git is not a new problem :smile:  The real problem is more we have too many solutions with differing tradeoffs/integration.

My 2c is that we need to consistently [generate RPMs in PRs](https://github.com/jlebon/redhat-ci/issues/10), *and* [from git master](https://github.com/cgwalters/rpmdistro-gitoverlay/) for integration testing.

Since we already have rhci hooked up for docker, I'd say the next simple step is to do an RPM build from the spec (clone fedora dist-git), until such time as this is more formalized in rhci or a successor.

cevich:
@cgwalters Yes, this needs to happen also, and redhat-ci is so simple, even an engineer can manage it 😀 
[We have separate cards for that work, including running the/some unittests.](https://trello.com/c/9dK8rNI4/384-8-cn-fiat-run-unittests-in-projectatomic-docker-prs)

I'm all about not re-inventing wheels, please tell us more about ``rpmdistro-gitoverlay`` and how we can/could use it here (high-level).

Oh, also, anyone that wants write access to my fork (the source of this PR), just say so.  No sense in forcingy'all to make pull-requests against my repo. to update this PR 😀

cevich:
> please tell us more about rpmdistro-gitoverlay

Got up off my lazy butt and did what the README is titled 😀

For the benefit of any other lazy-butts, rpmdistro-gitoverlay seems to solve the problem currently addressed via a mish-mash of ansible and bash scripting.  Using it here, does seem it would make things simpler.

miabbott:
> For the benefit of any other lazy-butts, rpmdistro-gitoverlay seems to solve the problem currently addressed via a mish-mash of ansible and bash scripting. Using it here, does seem it would make things simpler.

It's been used quite successfully in the [CentOS Atomic Host Continuous](https://wiki.centos.org/SpecialInterestGroup/Atomic/Devel) and [Fedora Atomic Host Continuous](https://pagure.io/fedora-atomic-host-continuous) streams, so there is some precedence.

miabbott:
Having a look at all the files now...looks like the bash scripts are duplicated under `tests/docker-rpm/` and `tests/docker-rpm/pkgs/`.  Is this intentional?

Also, with the inclusion of the bash scripts to do the building, is the intent for the test to build the RPMs, install them, and then test them?

cevich:
> so there is some precedence.

Thanks, I'm glad to hear there are existing users.  The references are handy, (@lsm5 usage hints).  I'll check out them out too.

>  Is this intentional?

Near as I can tell, @lsm5 scripts do more/less what we can do with Colin's tool and playbooks (in a more modular) way.  So mostly they're there for reference for now and will get rebased away after the useful bits are extracted/converted/transmogrified.

 >is the intent for the test to build the RPMs, install them, and then test them?

Yes that, but also to setup a reference and modules (roles) that can be exploited in other tests here and/or externally.

e.g. testing PRs.  This way, if the official, high-level integration tests pass (same test code and execution logic) but the low-level PR tests fail.  Logically, either something in the RPM building or the PR itself is responsible.  Since the details/workflow for both are common and accessible (simple), nobody has to go looking very far.

miabbott:
@cevich Is there additional work that needs to happen on this PR?  Or do you want to get merged?

cevich:
@miabbott Oh yes, lots probably, NO DO NOT MERGE.  @lsm5 is tasked with getting the actual building of RPMs working here.  For the moment, it's low on many priority lists (and should be on yours too 😀 )

miabbott:
Just slapping on some labels to make the status jump out a bit more when scanning PRs.

(Didn't want you to think this will never be merged, just not right now)

cevich:
Closing this as it's the wrong place for it given the [new proposal](https://fedoraproject.org/wiki/Changes/InvokingTests#Proposals_and_Evaluation).  Since this is "new" work, better to try and conform to that standard .  I'll move my code to another repo. for now, and if it bears any fruit (as a POC) it can be moved somewhere more somewhere more appropriate later on (like dist-git).

cevich:
New (Red Hat internal) temporary home: https://hurl.corp.redhat.com/406e7c6

PR #153, opened 2017-05-23
Support passing extra options to sub. man.

cevich:
Sometimes additonal options are needed during registration, such as
``--force`` or ``--org``.  Support this with an optional 'extra' key in
the CSV file.  When this key is not present, treat it as an empty
string to the sub. man. command-line.

### DIFF
--- a/roles/redhat_subscription/tasks/main.yml
+++ b/roles/redhat_subscription/tasks/main.yml
@@ -43,6 +43,7 @@
       --serverurl="{{ lookup('csvfile', 'serverurl file={} delimiter=,'.format(subscription_file)) }}"
       --username="$USERNAME"
       --password="$PASSWORD"
+      {{ lookup('csvfile', 'extra file={} delimiter=,'.format(subscription_file)) | default('') }}
     environment:
       USERNAME: "{{ lookup('csvfile', 'username file={} delimiter=,'.format(subscription_file)) }}"
       PASSWORD: "{{ lookup('csvfile', 'password file={} delimiter=,'.format(subscription_file)) }}"
--- a/roles/redhat_subscription/files/subscription_data.csv.sample
+++ b/roles/redhat_subscription/files/subscription_data.csv.sample
@@ -2,3 +2,4 @@ baseurl,https://cdn.redhat.com
 serverurl,subscription.rhn.redhat.com:443/subscription
 username,username@example.com
 password,yourPa$sW0rd
+extra,--force


cevich:
**N/B:** Totally untested.  Does this seem like a reasonable addition? If so, I'll carry it further.

miabbott:
Makes sense.  The default to empty string is a good move, too.

mike-nguyen:
Makes sense to me also.

cevich:
Kk thx, I'll take it to the next level then...

cevich:
One small (sanity) addition (better fix?).  Otherwise I tested this with a ``--org=123456`` and it works as expected.

miabbott:
> One small (sanity) addition (better fix?).

It looks like you force pushed a new commit, so the old changes aren't around for comparison any more.  This makes me sad.  ☹️ 

Please use `git commit --fixup` in the future or just push another commit.  Preserving the history during review is helpful, even if we tend to squash everything before a merge.  (I'll make a note of this preference in our [CONTRIBUTING.md](https://github.com/projectatomic/atomic-host-tests/blob/master/CONTRIBUTING.md) doc)

That being said, the changes look OK.

Huh, I just noticed:

> Sometimes additonal options are needed during registration, such as  --force or --org. 

We actually default to using `--force`.  Perhaps we can rework the options to maintain our default of always using `--force`, but give folks the option to not using it?  Just thinking aloud here.

cevich:
>  This makes me sad.

Oh noes!  Ya, I rebased/squashed.  NP, if you prefer to squash on merge, I can work with that.

We actually default to using --force.

Ohhhh, this is my fault.  I originally stuck that into the example CSV file, then later killed it (mistakenly) thinking it wasn't the current use.  Easy fix (good catch)...

roles/redhat_subscription/tasks/main.yml
```diff
@@ -43,6 +43,8 @@
       --serverurl="{{ lookup('csvfile', 'serverurl file={} delimiter=,'.format(subscription_file)) }}"
       --username="{{ lookup('csvfile', 'username file={} delimiter=,'.format(subscription_file)) }}"
       --password="{{ lookup('csvfile', 'password file={} delimiter=,'.format(subscription_file)) }}"
+      {{ lookup('csvfile', 'extra file={} delimiter=,'.format(subscription_file)) | default('') }}
+    # "fix vim's quote-counting
```
cevich: Do you have this problem too?  Is this comment here okay as a workaround?

miabbott: Huh, looks like I do.  No big deal with the comment.

cevich:
Update:  Rebased against master.  Removed vim-workaround comment (not needed anymore).

PR #157, opened 2017-05-24
Support subscribing RHEL hosts (no ostree)

cevich:
RHEL cloud images don't have ``/usr/bin/ostree`` installed
out-of-the-box.  Installing it, requires a subscription.
Support subscribing RHEL hosts by bypassing ostree dependant
tasks when ``/usr/bin/ostree`` is not present.


cevich:
**N/B:**  Tested on RHEL (with dd7a081ee4c212ba6c5f235c3c38396f30505d4f applied) and this works fine.  I'm not married to any variable names or using ``block``, happy to change either/both if desired.

miabbott:
This brings up an interesting point of discussion - what other roles will require changes to support non-Atomic Host systems?  As we expand our test coverage to things like `atomic` and `docker`, it makes sense to be able to run them on those non-Atomic Hosts.  Hmm...

I like the use of `block` here - no complaints.

roles/redhat_subscription/tasks/main.yml
```diff
@@ -21,14 +21,22 @@
...
-#
-  - name: Get origin file
-    command: ostree admin --print-current-dir
-    register: origin_file
 
-  - name: Get refspec from origin file
-    command: grep refspec {{ origin_file.stdout }}.origin
-    register: refspec
+  - name: status of /usr/bin/ostree is known
+    shell: type -Pf /usr/bin/ostree | /bin/true
```
miabbott: This could actually give a false positive if a host has `atomic` installed, which I believe pulls in `ostree` as a dependency.  Even if the dependency is not there, some host *could* have `ostree` installed and could cause the false positive.

You should probably check for `/run/ostree-booted` which *should* only be present on Atomic Host systems.  See https://github.com/projectatomic/atomic-host-tests/blob/master/roles/atomic_host_check/tasks/main.yml#L9

cevich: This is mostly due to my own ignorance, but...is there any harm possible running the affected commands on a non-atomic host, where ostree is present?  Related, is there any benefit (testing or otherwise) to running refspec/origin related tasks?

Another thing to consider is...maybe this role is too tightly-coupled to atomic-host and should be broken down into more discrete roles.  Again, I'm totally ignorant when it comes to how these refspec/origin tasks relate to sub. man.

Alternatively, I/we could split the tasks down into (local/static) include files.  e.g.:

```YAML
- name: Atomic-specific stuff happens before sub. man. command
  include: '{{ role_dir }}/tasks/atomic_stuff_before.yml
  when: is_atomic  # should probably set this in common somewhere

- name: Register with subscription-manager
  ...

- name: Atomic-specific stuff happens after sub. man. command
  include: '{{ role_dir }}/tasks/atomic_stuff_after.yml
  when: is_atomic  # should probably set this in common somewhere
```

cevich:
> This brings up an interesting point of discussion - what other roles will require changes to support non-Atomic Host systems?

I'm sure there are some, though based on some peeking, most of what's here seems very atomic/docker specific. In the interest of preventing role-sprawl (i.e. fifty different sub. man. roles), we should encourrage opening speculative PRs (like this one, or even less initially-functional) that:

* Try to re-use existing modules, possibly with changes, rather than adding new ones.
* Try not to (intentionally) break any existing playbooks with behavior changes.
* Present a case-by-case platform for discussion and consensus on how to proceed.

roles/redhat_subscription/tasks/main.yml
miabbott: Well, trying that particular command on a F25 Cloud system blows up and returns `1`, so that would cause the role to fail and kill the whole playbook.

The role is definitely coupled to Atomic Host, because we didn't consider running these tests on anything but an Atomic Host.  😎 

I believe the reason we are manipulating the `.origin` file here is actually a side effect of how we provision systems in our Jenkins jobs.  Without getting into those details, when a system gets registered via sub-man, that refspec can get clobbered.  @mike-nguyen can correct me here, if I'm wrong.

In this case, I'm not in favor of splitting out the Atomic-specific stuff because it would amount to only a task or two per role.

cevich:
Example: https://github.com/projectatomic/atomic-host-tests/pull/160

cevich:
Re-based, added ``fixup!`` as an option for review (untested).  TBH, I'm not keen on role-dependencies ``meta/main.yml`` (they're kind of TOO-magic).  Maybe there's a better way?

roles/redhat_subscription/tasks/main.yml
cevich: Okay, blocks are no problem either.  What about the idea of re-using ``atomic_host_check`` (for ``is_atomic``)?

I don't needing ``meta/main.yml``, maybe it can be worked the other way around though...

Stick the atomic-detection task under ``common`` then include that here and in ``atomic_host_check``?  (and I can throw away #160)

cevich:
Okay, this version drops the small include files in favor of using a block, and adds re-use of the ``is_atomic`` code by this and the ``atomic_host_check`` role.  With this, we can optionally/simply drop #160 and everything will continue to work as it was before.

cevich:
All done, that's more than plenty for one day 😀

I'm happy to combine some of these little PRs under one-roof if that's easier (was trying to keep them each discrete units).  No rushes...sky's not falling for any of these.

miabbott:
bot, retest this please

cevich:
(rebase & force-push)

roles/atomic_host_check/tasks/main.yml
```diff
@@ -6,21 +6,9 @@
...
-- name: Init the is_atomic fact
-  set_fact:
-    is_atomic: false
-
-- name: Set the is_atomic fact
-  set_fact:
-    is_atomic: true
-  when: s.stat.exists
+- name: Determine is_atomic True/False
+  include: '{{ playbook_dir }}/common/set_is_atomic.yml'
```
miabbott: Hooray for CI!  Our `redhat-ci` checks found this problem:

```
TASK [atomic_host_check : Determine is_atomic True/False] **********************
    "failed": true, 
    "reason": "the file_name '/var/tmp/checkout/tests/improved-sanity-test/common/set_is_atomic.yml' does not exist, or is not readable"
}
```

Looks like the `include:` statement is pointing at the wrong place.

cevich:
investigating...

cevich:
Heh, you beat me to it 😀 (we cross-posted).  Okay, fixing...

cevich:
Suggestion:  Consider (long-term) putting ``common`` under roles?  That way it's relative position isn't dependent at all on the directory the playbook is running from.  i.e. move it to ``roles/common``.

Another solution is to always/additionally symlink ``common`` to ``{{ playbook_dir }}`` (same as how you do for ``roles``), then relative references as ``{{ playbook_dir }}/common`` will always work for includes, just as ``{{ playbook_dir }}/roles/whatever/tasks/main.yml`` (used in a few places IIRC).

roles/redhat_subscription/tasks/main.yml
```diff
@@ -21,14 +21,21 @@
...
-#
-  - name: Get origin file
-    command: ostree admin --print-current-dir
-    register: origin_file
 
-  - name: Get refspec from origin file
-    command: grep refspec {{ origin_file.stdout }}.origin
-    register: refspec
+  - name: Determine is_atomic True/False
+    include: '{{ playbook_dir }}/../common/set_is_atomic.yml'
```
miabbott: Still failing:

```
TASK [atomic_host_check : Determine is_atomic True/False] **********************
    "failed": true, 
    "reason": "the file_name '/var/tmp/checkout/tests/common/set_is_atomic.yml' does not exist, or is not readable"
}
```

Try `include: ../../common/set_is_atomic.yml`

It's not the best solution, but would work in the short-term

cevich:
> I would like to get rid of common, just not a priority at the moment.

I understand, just putting it out there to roll around in your skull 😀   Actually I've found including from a "library" like this works much better than ``include_role`` in Ansible.

mike-nguyen:
@cevich   I just looked up include_role and I'm just curious what specifically about including from a "library" works much better than include_role?  We have just been using the include statement.

cevich:
@mike-nguyen I believe it was an ansible bug in 2.something, because I can no-longer reproduce it under 2.3.  Basically, if you passed task-vars, ``include`` would represent them fine, but ``include_role`` would **only** represent those vars (i.e. no other hostvars were visable).  So:

```yaml
- include: "/path/to/file.yml"
  vars:
    foo: "bar"
```

vs

```yaml
- include_role: "baz"
  vars:
    foo: "bar"
```

in the second case, the *only* variable available was "foo". In the former, all the normal ``hostvars`` were available along with ``foo``.  Though it's possible I'm remembering the problem incorrectly.

miabbott:
I think this is safe to merge, however I'm wary of doing that on a Friday and having breakages over the weekend.

Will revisit this next week.

PR #158, opened 2017-05-24
Hide subscription-manager username/password

cevich:
It's almost always a good idea to protect in-flight credentials.  Wrap
the username & password values inside temporary environment variables.
These env. vars. only exist for the life of the ``bash`` process.
Though they will still be exposed in a process-listing (on the target),
they won't show up inside any ansible logs.

### DIFF
--- a/roles/redhat_subscription/tasks/main.yml
+++ b/roles/redhat_subscription/tasks/main.yml
@@ -41,8 +41,11 @@
       subscription-manager register --auto-attach --force
       --baseurl="{{ lookup('csvfile', 'baseurl file={} delimiter=,'.format(subscription_file)) }}"
       --serverurl="{{ lookup('csvfile', 'serverurl file={} delimiter=,'.format(subscription_file)) }}"
-      --username="{{ lookup('csvfile', 'username file={} delimiter=,'.format(subscription_file)) }}"
-      --password="{{ lookup('csvfile', 'password file={} delimiter=,'.format(subscription_file)) }}"
+      --username="$USERNAME"
+      --password="$PASSWORD"
+    environment:
+      USERNAME: "{{ lookup('csvfile', 'username file={} delimiter=,'.format(subscription_file)) }}"
+      PASSWORD: "{{ lookup('csvfile', 'password file={} delimiter=,'.format(subscription_file)) }}"
     register: result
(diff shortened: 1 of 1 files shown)


cevich:
**N/B:** Untested / speculative (but should work).  Worth pursuing?

miabbott:
We don't publicly publish our RHELAH test logs/results, so I think the risk of leaked credentials is low.

However, if you could confirm this works, this change would be A-OK.

mike-nguyen:
I'm onboard with this.  It is another layer of protection just in case the log is saved or copied somewhere public.  

I just tested it on RHELAH and it works fine.  Output is as follows:

TASK [redhat_subscription : Register with subscription-manager] ****************
changed: [rhelah] => {
    "attempts": 1, 
    "changed": true, 
    "cmd": [
        "subscription-manager", 
        "register", 
        "--auto-attach", 
        "--force", 
        "--baseurl=cdn.redhat.com", 
        "--serverurl=subscription.rhn.redhat.com:443/subscription", 
        "--username=$USERNAME", 
        "--password=$PASSWORD"
    ], 
    "delta": "0:00:19.603818", 
    "end": "2017-05-25 14:51:51.537359", 
    "rc": 0, 
    "start": "2017-05-25 14:51:31.933541", 
    "warnings": []
}

STDOUT:
---
Registering to: subscription.rhn.stage.redhat.com:443/subscription
The system has been registered with ID: b6e631f5-9e2b-4ba8-9d2d-b92e90be876d 

Installed Product Current Status:
Product Name: Red Hat Enterprise Linux Atomic Host
Status:       Subscribed

Product Name: Red Hat Enterprise Linux Server
Status:       Subscribed
---

cevich:
> We don't publicly publish our RHELAH test logs/results

They could easily leak via posting to bugzilla or maybe someone internally gets frustrated with something and needs to "borrow" some test-credentials...naa...nevermind, that would NEVER happen 😀 

Security and Onions go well together (especially with breath-mints).

cevich:
**Note:** There is a conflict here with #153 - I'm happy to resolve that either way, if/not this/that gets merged.

miabbott:
We have confirmation this works, so let's green light this.

PR #160, opened 2017-05-25
Support atomic_host_check !fail when !is_atomic

cevich:
Support all existing playbooks by adding an optional flag to shut-off
the fail task.  This allows the role to be re-used on all platforms, and
gives the caller the power to decide whether or not to enforce
verification-failure.

### DIFF
--- a/roles/atomic_host_check/defaults/main.yml
+++ b/roles/atomic_host_check/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+
+# Set false if role should NOT fail when is_atomic == False
+fail_if_not_atomic_host: True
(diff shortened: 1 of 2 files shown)


cevich:
**N/B:**  Untested / Speculative.

A (possibly prettier) implementation could be to separate the failure from the checking, but this would require modifying all existing playbooks 😦 

A compromise could be to add a new role, say "atomic_query", and just share (by ``include``) the checking bits.  But this adds a new role w/ possibly confusing/overlapping name 😦

cevich:
Closing in favor of the "includes" / "is_atomic" solution from  #157

PR #161, opened 2017-05-25
Adjust requirements.txt to stay at 2.2.x

miabbott:
In the previous version, `requirements.txt` would install Ansible 2.3.  Probably wouldn't break anything, but let's just stay paranoid and require users to stick at 2.2

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-ansible>=2.2.0
+ansible<2.3


requirements.txt
```diff
@@ -1 +1 @@
-ansible>=2.2.0
+ansible<2.3
```
cevich: FWIW, I think 2.3 is what's currently shipping in CentOS.  The little bits of tinkering I've done with it in ADEPT have been mostly positive.  Though IIRC it does alter some behaviors in non-backward compatible ways, the fixes needed were pretty minor.

miabbott: It might be a fool's errand, but I don't want to be in the position of having to continually make changes to support the latest release.

It looks like Ansible releases about every 4 months, but supports the last two major releases.

http://docs.ansible.com/ansible/dev_guide/developing_releases.html

I guess this means we'll have to re-evaluate which versions we support about every 8 months?

cevich:
```diff
diff --git a/.redhat-ci.yml b/.redhat-ci.yml
index 47a9078..88ebd24 100644
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -9,6 +9,7 @@ cluster:
 
 packages:
   - ansible
+  - rsync
 
 tests:
   - ansible-playbook -vi testnode, common/ans_ah_head-1_deploy.yml
```

?

cevich:
> I guess this means we'll have to re-evaluate which versions we support about every 8 months?

IMHO, you can go a lot longer than that, esp. if you stick to just the core modules and avoid doing fancy-stuff.  Lack of "ansible support" doesn't mean it suddenly stops working 😀   Just that nobody's going to backport fixes or try to find new bugs.  In many cases you end up finding workarounds anyway.  Unless you're desperate for some new feature, or simply can't get the old release anymore, then it's an easier argument to make.

miabbott:
bot, retest this please

miabbott:
@cevich I don't see the addition of `rsync` in my changes.  Are you suggesting that should be added or are you seeing it yourself?

mike-nguyen:
Lets go with just the Ansible change for now and if we need to add rsync later on, it is an easy fix.  Having 2.3 installed might cause some unexpected issues with our tests designed to run on 2.2.0.

PR #163, opened 2017-05-31
docs: howto contribute to the repo

miabbott:
It was requested that additional documentation be made available about
how to start contributing to the repo.  I've created a new `docs`
directory that currently holds a single file that goes into some of
the details of the repo and how to create roles/tests.

### DIFF
--- a/docs/howto_contribute.md
+++ b/docs/howto_contribute.md
@@ -1,0 +1,309 @@
+### A Brief Introduction to the Repo
+
+At the top-level of the repository, we have five directories.  Of these five
+directories, the `roles` and `tests` directories are where the majority of
+the development happens.
+
+The `callback_plugins` directory is where we have a single plugin that
+provides improved output formatting from the Ansible playbooks.  It also
+provides some functionality that allows the tests to collect the journal
+from a system after a failure.
+
+The `common` directory contains pieces of Ansible playbooks/roles that did
+not fit in with the other roles and tests that exist, but are still used
+in parts of the repository.
+
+The `docs` directory contains documentation important to using the repo
+and developing for the repo.
+
+The `roles` directory is the traditional `roles` directory that is often
+used in Ansible playbooks.  We have placed the directory at the top-level
+of the repo, so that we can re-use the roles across multiple tests.
+
+The `tests` directory contains playbooks which execute tests using a mix
+of roles and tasks.
+
+### Writing Your First Role
+
+When writing a role for to be used in a test, you should attempt to make it
+fairly generic, so that it can be re-used by other users.  To show this, let's
+look at an existing role which is **NOT** generic and could be modified.
+
+The `docker_build_httpd` role was made to support building a specific Docker
+image.  It is shown in its entirety below (from [roles/docker_build_httpd](/roles/docker_build_httpd/tasks/main.yml)):
+
+```yaml
+---
+# vim: set ft=ansible:
+#
+- name: Fail if g_osname not set
+  fail:
+    msg: "The g_osname variable is not defined"
+  when: g_osname is not defined
+
+- name: Fail if g_httpd_name not set
+  fail:
+    msg: "The g_httpd_name variable is not defined"
+  when: g_httpd_name is not defined
+
+- name: Create temp directory for building
+  command: mktemp -d
+  register: m
+
+- name: Set build_dir fact
+  set_fact:
+    build_dir: "{{ m.stdout }}"
+
+- name: Copy Dockerfile to temp directory
+  copy:
+    src: "{{ g_httpd_name }}_Dockerfile"
+    dest: "{{ build_dir }}"
+    owner: root
+    group: root
+    mode: 0644
+
+- name: Copy makecache.sh to temp directory
+  copy:
+    src: makecache.sh
+    dest: "{{ build_dir }}"
+    owner: root
+    group: root
+    mode: 0744
+  when: ansible_distribution != 'RedHat'
+
+- name: Copy rhel_makecache.sh to temp directory
+  copy:
+    src: rhel_makecache.sh
+    dest: "{{ build_dir }}"
+    owner: root
+    group: root
+    mode: 0744
+  when: ansible_distribution == 'RedHat'
+
+- name: Build httpd image
+  command: "docker build -t {{ g_httpd_name }} -f {{ build_dir }}/{{ g_httpd_name }}_Dockerfile {{ build_dir }}"
+
+- name: Get docker images after build
+  command: docker images
+  register: build_images
+
+- name: Fail if httpd image not present
+  fail:
+    msg: "The {{ g_httpd_name }} image is not present"
+  when: "'{{ g_httpd_name }}' not in build_images.stdout"
+```
+
+In this example, the role makes an assumption about the name of the Dockerfile,
+the existence of a `makecache.sh` script, and relies on a global variable
+throughout.
+
+This role could be improved by removing the assumption about the `makecache.sh`
+script, parameterizing more of the tasks, and separating out tasks that
+manipulate files/directories into separate roles/tasks.  The improved version
+should focus on just the `docker build` command.
+
+An improved version would live in `roles/docker_build/tasks/main.yml` and
+could look like this:
+
+```yaml
+---
+- name: Fail if required variables are not set
+  fail:
+    msg: "The required variables are not set."
+  when: image_name is not defined or
+        build_path is not defined or
+        dockerfile_name is not defined
+
+- name: Build Docker image
+  command: "docker build -t {{ image_name }} -f {{ build_path }}/{{ dockerfile_name }} {{ build_path }}"
+
+- name: Verify Docker image was built
+  command: docker images
+  register: docker_images
+  failed_when: "'{{ image_name }}' not in docker_images.stdout"
+```
+
+This completely simplifies the role to handle the building of a Docker image
+when given the required parameters.  In this simplified version, you would
+have to have separate roles/tasks that would handle the creation of any
+required directories and copying the Dockerfile to the host.
+
+### Writing Your First Test
+
+Now you have a new role created, let's put it to use in a new test playbook.
+
+#### Directory Setup
+
+Firstly, you will have to create a new sub-directory under `tests` and create
+symlinks to the top-level `callback_plugins` and `roles` directories.  This
+allows the test to reference the available roles and have its output nicely
+formatted by the callback plugin.  In this example, we are going to pretend we
+are creating a test playbook for the `docker build` command.
+
+```bash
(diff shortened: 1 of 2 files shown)


miabbott:
@ddarrah Have a look please

And anyone else, too.  I didn't really know what I should cover vs. what was unnecessary, so I'm happy to edit this until satisfaction is reached.

miabbott:
[(2) Document how to contribute to the [a-h-t] repo](https://trello.com/c/8gc4aAQ9/730-2-document-how-to-contribute-to-the-a-h-t-repo)

mike-nguyen:
Looks great @miabbott!  Maybe we can add a section at the end for common gotchas.  I can't think of anything off the top of my head at the moment but I feel like I run into issues that we have worked around.  We can add this later if the same questions or issues are being asked.

cevich:
Maybe add (I haven't looked at the document):
*  Recommended local git setup
*  Recommended ``vim`` setup (i.e. ansible plugin, where to get it, how to install, etc...**cough**emacs**cough**

cevich:
@miabbott another place I remember there's a good reference for this stuff is actually the (very old) [autotest developer docs](http://autotest.readthedocs.io/en/latest/main/developer/index.html).  At the time, that project was in a very similar situation, w/ lots of folks from different parts of the world.  Esp. that "code submission checklist" was particularly successful, because it gave a very quick way for a reviewer to request changes to a PR w/o spending much time explaining.  Just point and say, you need to do ``<checklist item>``.

miabbott:
FYI, if you want to see the doc rendered, you can have a look over in my branch - https://github.com/miabbott/atomic-host-tests/blob/contributing_to_aht/docs/howto_contribute.md

miabbott:
>     Recommended local git setup

I don't think we are doing anything special here.  I'm making the assumption that folks have used `git` and GitHub before.  But I could probably note that assumption in the docs.

>     Recommended vim setup (i.e. ansible plugin, where to get it, how to install, etc...coughemacscough

I'll note how to setup `vim` for Ansible, but also mention it is not required to use `vim`.

>     Also, best spell out tab/space and general coding standards as well.

This is partially covered in the [CONTRIBUTING.md](https://github.com/projectatomic/atomic-host-tests/blob/master/CONTRIBUTING.md) where we link out to the OpenShift Style Guide + Best Practices.  However, I'll add another note about that in this doc, too.

cevich:
> I'll note how to setup vim for Ansible, but also mention it is not required to use vim.

Related: might mention about having those ``# vim: set ft=ansible:`` lines in every YAML 😄

miabbott:
New changes based on feedback ⬆️

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+This `vim` plugin is also why many of the `.yml` files in the repo have the
+following comment at the head of file:
+
+`# vim: set ft=ansible:`
+
+### Writing Your First Role
+
+When writing a role for to be used in a test, you should attempt to make it
+fairly generic, so that it can be re-used by other users.  To show this, let's
+look at an existing role which is **NOT** generic and could be modified.
```
cevich: This is a pretty broad topic, getting the right amount of flexibility and yet avoiding excessive complication.  Use of example is good in general, but I'm generally not a fan of negative-examples.  Generally-speaking, there's an infinite number of ways to write bad code, but fewer ways to write bad-but-acceptable code :D   IMHO, it's better to spend the document-space outlining the qualities you're looking for.

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+as it does not cause any breakage in the roles/tests or distrupts the overall
+style of the repo.
+
+That being said, `vim` users would probably be interested in the the
+[Ansible plugin](https://github.com/pearofducks/ansible-vim) which enables syntax highlighting.
+This `vim` plugin is also why many of the `.yml` files in the repo have the
+following comment at the head of file:
+
+`# vim: set ft=ansible:`
+
```
cevich: Great so far, I really like how you've explained all the basic concepts above without getting too detailed.

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+Firstly, you will have to create a new sub-directory under `tests` and create
+symlinks to the top-level `callback_plugins` and `roles` directories.  This
+allows the test to reference the available roles and have its output nicely
+formatted by the callback plugin.  In this example, we are going to pretend we
+are creating a test playbook for the `docker build` command.
+
+```bash
+$ mkdir -p tests/docker-build/
+$ ln -s ../../callback_plugins/ tests/docker-build/callback_plugins
+$ ln -s ../../roles/ tests/docker-build/roles
```
cevich: This is probably fine for starting out, but...something to think about:

```bash
mkdir -p tests/foobar/roles
cd tests/foobar/roles
ln -s ../../../roles/* .
```

Allows writing new/experimental test-local roles w/o polluting the collective namespace.  It also allows excluding unnecessary shared-roles (just delete the links) that aren't used, keeping the local role namespace simple.

Totally optional, and maybe not so good for a "getting started" type guide.  Just something to think about.

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+Additional directories such as `files`, `templates` or `vars` may be required
+for your test.  Please reference the [Ansible documentation](http://docs.ansible.com/ansible/playbooks_best_practices.html#content-organization) about
+content organization about when and how to use these directories.
+
```
cevich: Maybe stick a quick blurb in here about setting up ``redhat_subscription/files/subscription_data.csv.sample``?

Which reminds me, that may be a good thing to add to your ``.gitignore`` file, ``subscription_data.csv``, good to prevent obvious woopsies where you can.

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+where each set of features is separated into their own playbooks.
+
+For our example, we will use the simple case of a single playbook in a single
+file.
+
+#### Playbook Sections
+
+All of the playbooks should start with a `name`, `hosts`, and `become`
+declaration.  The name is up to the test developer, but we almost always use
+`hosts: all` and `become: yes`.  The `hosts: all` allows the executor of the
```
cevich: This is your child, so do what you know/like.  My experience has been that ``hosts: all`` can end up (in the long-run) causing a major headache in cases where you want to include one play in another, and/or don't want all hosts to receive all roles.  OTOH, using groups does bring inventory-aspects into your plays more than maybe you'd like, I get that.

However, Ansible does allow you to have it both ways by using a pattern-variable, like:

```YAML
- hosts: "{{ pattern_variable | default('all') }}"
```

or something to that effect...where the actual name of ``pattern_variable`` doesn't matter so much, as long as it's unique per-play.  This supports future growth and flexibility through increased playbook-modularity.  Enabling re-use of plays, and preventing the friction of wanting to re-use plays later, but not being able to because of the extensive changes required.

Anyway...just something to ponder, maybe it's not important here.

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+
+Often a tests will require some tasks to be performed that are not covered
+by any existing roles.  In this situation, one would use the `pre_tasks`
+statement to define the tasks that should be executed ahead of the roles.
+
+After any required `pre_tasks` the test should be built using the roles
+available in the top-level directory.
+
+Finally, any tasks that need to be run after the roles can be specified
+in the `post_tasks` section.
```
cevich: I wouldn't even mention ``pre_tasks`` and ``post_tasks``.  You want to encourage writing re-usable roles.  There should be very few, and very specific cases where you need to use pre/post tasks (like installing python on Fedora before running setup).  I say drop this section, it's not something newcomers should have to worry about and even if so, you already stated "familiarity with Ansible" in the requirements above, so it's okay to leave this in that :D

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### Example Playbook
+
+In the example playbook, we are going to copy a couple of Dockerfiles to the
```
cevich: I would suggest inserting a step #0 in here:  Write a README.md file in the new tests directory :D  The effect is, it makes the author consider the big-picture of what they're trying to accomplish before writing any code.  That almost always results in better quality code AND documentation.  Otherwise the docs. are always an afterthought (or omitted all together).

docs/howto_contribute.md
```diff
@@ -0,0 +1,396 @@
...
+which may interfere with other tests that are run afterwards.
+
+### Testing Your Playbook
+
+After you have finished with your playbook, it is important to test that it can
+successfully run against the Atomic Host variant of CentOS, Fedora, and RHEL.
+If one of those platforms does not run successfully, you'll need to make the
+necessary adjustments before submitting a pull request with your changes.
+
+### Documenting Your Playbook
```
cevich: Ahh here it is.  Suggest moving this up, as the **first** item.  Emphasizing the big-picture planning and docs maintence, rather than leave it to an after thought.  Most people will read the example playbook section (above), and stop there if accidentally works :S

docs/howto_contribute.md
miabbott: Will note this.  Our `.gitignore` already has `*.csv`, so we should be covered.

docs/howto_contribute.md
miabbott: The `default()` filter is a good suggestion.  We've been able to get away with `hosts: all` thusfar because the focus has been single node tests.

I'll investigate updating our `hosts:` declaration across our tests and see how it works for us.

docs/howto_contribute.md
miabbott: Good suggestion; I'll move this up in the docs.

docs/howto_contribute.md
miabbott: Yeah, I think I can highlight what we are looking for with the just the 'fixed example' and not bothering to show the 'wrong example'.

docs/howto_contribute.md
miabbott: The drawback to this is that you will end up with (currently) 68 symlinks to decide to delete.  And then after your pruning is complete, the git commit will have a number of symlinks included (functionally not a problem, but aesthetically displeasing, in my opinion).

cevich: Yep, np.

docs/howto_contribute.md
cevich: oops, my bad.

miabbott:
@cevich Thanks for all the feedback.  I've incorporated some of your ideas into the most recent changes ⬆️

miabbott:
If you have additional feedback, please leave it here in the next 24 hours.  Otherwise, I'm going to merge this tomorrow.

PR #164, opened 2017-05-31
Add rpm-ostree tests

mike-nguyen:
Add a basic test suite for rpm-ostree.  There are still a few subcommands
that still need to be added to the test suite.  These tests assume
that the atomic host is provisioned with the latest version as it will
move between HEAD and HEAD-1.

Tests:
- Use rpm-ostree to deploy by version then rollback and delete rollback
- Use rpm-ostree to deploy by checksum then rollback and delete rollback
- Use rpm-ostree to deploy HEAD-1 then use cleanup pending deployments
- Simulate rpm-ostree upgrade then rebase to HEAD and delete rollback
- Simulate rpm-ostree upgrade with package install, then rebase to HEAD
  with an uninstall of the same package, then delete rollback
- Enable and validate client side initramfs generation then disable
  initramfs and cleanup deployments


mike-nguyen:
I tested on CAHC, RHELAH 7.3.5, and Fedora 26.  The rpm-ostree version installed must support the --install and --uninstall flags so CentOS AH may not have it yet.

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+      command: rpm-ostree cleanup -rpmb
+
+#####################################################################################
+#
+# rpm-ostree initramfs
+#   - test client side initramfs
+#
+#####################################################################################
+
+- name: rpm-ostree - upgrade and rebase + install and uninstall
```
jlebon: Looks like this needs to be changed? :)

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+      command: test -n {{ initramfs.stdout }}
+
+    - name: Get contents of initramfs
+      command: lsinitrd {{ initramfs.stdout }}
+      register: lsinitrd
+
+    - name: Fail if contents of initramfs is incorrect
+      fail:
+        msg: "Expected /etc/rpmostree-initramfs-testing-file in initramfs"
+      when: "'/etc/rpmostree-initramfs-testing-file' not in lsinitrd.stdout"
```
jlebon: There's a subtle issue with this. The output of `lsinitrd` also prints out the arguments that were used to create the initramfs, which I suspect why this is matching. (The actual file listings don't have a leading `/` in their filenames).

Maybe a better check here is `lsinitrd $initrd -f /my/file` and checking that e.g. `This is the contents of my file` is in the output?

tests/rpm-ostree/README.md
```diff
@@ -0,0 +1,45 @@
...
+  - rpm-ostree initramfs
+  - rpm-ostree install
+  - rpm-ostree uninstall
+
+To be Covered
+  - rpm-ostree db
+
+Covered In Another Test
+  - rpm-ostree install
+  - rpm-ostree uninstall
```
mike-nguyen: The package layering feature has a separate test suite already in atomic-host-tests.  The install/uninstall below is the package layering as options to deploy, upgrade, and rebase

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+    - name: Convert to json
+      set_fact:
+         ross_json: "{{ ros_status.stdout | from_json }}"
+
+    - name: Set current version
+      set_fact:
+         head_version: "{{ ross_json['deployments'][0]['version'] if ross_json['deployments'][0]['booted'] else ross_json['deployments'][1]['version'] }}"
+
+    - name: Pull last two commits
+      command: ostree pull --subpath /usr/share/rpm --depth=1 {{ refspec.stdout }}
```
mike-nguyen: We switched to this for  CAHC and FAHC.  I was away when this change happened so I don't know the specifics.  It has something to do with commit history on the continuous streams.  @miabbott do you know offhand why we moved to this instead of commit-metadata-only?

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+
+    - name: Set current commit version
+      set_fact:
+         head_csum: "{{ ross_json['deployments'][0]['checksum'] if ross_json['deployments'][0]['booted'] else ross_json['deployments'][1]['checksum'] }}"
+
+    - name: Create local branch
+      command: ostree refs --create local-branch {{ head_csum }}
+
+    - name: Update origin file
+      command:  sed -i 's/^\(.*refspec\)=.*$/\1=local-branch/g' {{ current_dir.stdout }}.origin
```
mike-nguyen: I'll add the rpm-ostree reload then verify the rpm-ostree status output which is a test for rpm-ostree reload.

tests/rpm-ostree/main.yml
miabbott: Ah, we started doing this in our deployment playbooks/ jobs:

```
commit 7857fe3
Author: Micah Abbott <miabbott@redhat.com>
Date:   Wed Oct 12 11:01:25 2016 -0400

    ans_atomic_info: fetch commit and file/RPM data
    
    This changes the playbook to pull the file/RPM data for use in determining
    the the package differences between HEAD^ and HEAD of the stream being
    tested.
    
    We had avoided doing this in the past because an `ostree pull
    --commit-metatdata-only` of the CAHC stream would fetch the deltas
    being generated between the two commits.  Thus, the `rpm-ostree db
    diff` would return the desired results.
...
```

I think in this instance we are safe just pulling the commit metadata

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+
+  tags:
+    - deploy_version
+
+  vars_files:
+    - vars.yml
+
+  tasks:
+    - name: Get current refspec
+      shell: cat $(ostree admin --print-current-dir).origin | grep refspec | cut -d= -f2
```
miabbott: Can we get this from the JSON output of `rpm-ostree status`?  It looks like it may be the `origin` value:

```
# rpm-ostree status --json | python -m json.tool
{
    "deployments": [
        {
            "booted": true,
            "checksum": "0ccf9138962e5c2c3794969a228e751d13bb780f5b0a1f15f4a9649df06ba80a",
            "gpg-enabled": false,
            "id": "rhel-atomic-host-0ccf9138962e5c2c3794969a228e751d13bb780f5b0a1f15f4a9649df06ba80a.0",
            "origin": "rhel-atomic-host:rhel-atomic-host/7/x86_64/standard",
            "osname": "rhel-atomic-host",
            "packages": [],
            "regenerate-initramfs": false,
            "requested-local-packages": [],
            "requested-packages": [],
            "serial": 0,
            "timestamp": 1495490444,
            "unlocked": "none",
            "version": "7.3.5"
        }
    ],
    "transaction": null
}
```

@jlebon Can you confirm this?

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+
+    - name: Pull last two commits
+      command: ostree pull --subpath /usr/share/rpm --depth=1 {{ refspec.stdout }}
+      register: osp
+      until: osp.rc == 0
+      retries: 5
+      delay: 5
+
+    - name: Get HEAD-1 checksum
+      shell: ostree log {{ refspec.stdout }} | grep ^commit | tail -1 | cut -d' ' -f2
```
miabbott: You can get this with `ostree rev-parse $refspec^`.  Like this:

```
# ostree rev-parse rhel-atomic-host/7/x86_64/standard^
d6c7a5639cdeb6c21cf40d80259d516d047176e35411c8684cae40a93eedbed0
```

roles/rpm_ostree_rebase/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+# vim: set ft=ansible:
+#
+# rpm-ostree rebase
+#   params:
+#     remote_name (required) name for remote
+#     remote_url (required) remote url
+#     refspec (required) remote refspec
+
```
miabbott: Do you want to add a `fail` statement here if the parameters aren't defined?

miabbott:
This is a good set of tests.  I'd echo what has been said already that if we could make a role that basically generates a variable containing the JSON of `rpm-ostree status`, it would cleanup the code duplication.

tests/rpm-ostree/README.md
mike-nguyen: Just read this again.  I will remove it form the readme.

tests/rpm-ostree/main.yml
mike-nguyen: Great!  I made the changes to commit metadata

tests/rpm-ostree/main.yml
jlebon: Yup, that's correct! I made the same suggestion a bit lower down on another instance of this.

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,571 @@
...
+
+    - include: ../../common/ans_reboot.yml
+
+    - name: Get rpm-ostree status output
+      command: rpm-ostree status --json
+      register: ros_status
+
+    - name: Convert to json
+      set_fact:
+         ross_json: "{{ ros_status.stdout | from_json }}"
```
mike-nguyen: Yes, it works.  There was already the rpm_ostree_status role that converts to json then sets ross_json and also sets booted and unbooted to the proper deployment.

mike-nguyen:
Pushed fixup with the changes suggested.

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+      register: osp
+      until: osp.rc == 0
+      retries: 5
+      delay: 5
+
+    - name: Get HEAD-1 version
+      shell: ostree show $(ostree rev-parse {{ refspec }}^) --print-metadata-key version | tr -d \'
+      register: hmo_version
+
+    - name: Deploy version {{ hmo_version.stdout }}
```
miabbott: nit - I don't think variable substitution happens for the `name:` statements

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+      register: ros_deploy
+      until: ros_deploy.rc == 0
+      retries: 5
+      delay: 5
+
+    # reboot
+    - include: ../../common/ans_reboot.yml
+
+    # verify version in deployment 0
+    - include: roles/rpm_ostree_status_verify/tasks/main.yml
```
miabbott: Is there a reason to mix `include` vs. `include_role` in this test?

roles/rpm_ostree_status_verify/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+       Expected: {{ num_deployments }}.
+       Actual {{ ros_num_deployments }}"
+  when: num_deployments is defined and {{ num_deployments | int }} != {{ ros_num_deployments | int }}
+
+- name: Fail if deployment does not exist
+  fail:
+    msg: "deployment {{ deployment }} does not exist"
+  when: ros_json['deployments'][deployment] is undefined
+
+- name:  Fail if deployment {{ deployment }} values are incorrect
```
miabbott: I was a bit confused why you used a `dict` here, but seeing it used in practice in the main test file made me understand.

Great idea and very flexible!

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+  tags:
+    - deploy_version
+
+  vars_files:
+    - vars.yml
+
+  tasks:
+    - include: roles/rpm_ostree_status/tasks/main.yml
+
+    - name: Set current version and refspec
```
miabbott: Might be helpful to stick a comment here explaining that `ros_booted` comes from the `rpm_ostree_status` role that was included above.

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+         refspec: "{{ ros_booted['origin'] }}"
+
+    - name: Get origin dir
+      command: ostree admin --print-current-dir
+      register: current_dir
+
+    - name: Create local branch
+      command: ostree refs --create local-branch {{ head_csum }}
+
+    - name: Update origin file
```
miabbott: Maybe drop a comment here explaining why you are modifying the `origin` file

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+    # verify rebase back to original refspec
+    - include: roles/rpm_ostree_status_verify/tasks/main.yml
+      num_deployments: 2
+      deployment: 0
+      expected:
+        checksum: "{{ head_csum }}"
+
+    - name: Cleanup
+      command: rpm-ostree cleanup -rpmb
+
```
miabbott: Should there be `rpm_ostree_status_verify` here as in previous sections?

tests/rpm-ostree/main.yml
mike-nguyen: Seems like it does although I have seen instances when it doesn't (possibly in loops?)

```
TASK [Deploy version 7.2017.383] ***********************************************
changed: [fedora] => {
    "attempts": 1, 
    "changed": true, 
    "cmd": [
        "rpm-ostree", 
        "deploy", 
        "26.55"
    ], 
    "delta": "0:00:32.515279", 
    "end": "2017-06-06 14:46:03.710649", 
    "rc": 0, 
    "start": "2017-06-06 14:45:31.195370", 
    "warnings": []
}
```

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+    - include: roles/rpm_ostree_status_verify/tasks/main.yml
+      num_deployments: 2
+      deployment: 0
+      expected:
+        booted: true
+        checksum: "{{ head_csum }}"
+
+    - name: Cleanup
+      command: rpm-ostree cleanup -rpmb
+
```
miabbott: Another instance to verify that the cleanup was successful with `rpm_ostree_status_verify`?

roles/rpm_ostree_status_verify/tasks/main.yml
mike-nguyen: My original implementation had dynamically named variables based on the rpm-ostree status json keys.  The problem was that certain properties are displayed based on some rpm-ostree actions.  If an action is undone, sometimes that property would disappear from the output but the variable would persist and cause some confusion validating property values.

tests/rpm-ostree/main.yml
mike-nguyen: I was kind of all over the place with the include vs include role.  I was trying to use include_role in as many places as possible because it takes out the relative path.  The instances I did not was:

1. ans_reboot.yml isn't in a roles directory so it needs to be included
2. variables not being passed properly using include_role.  I think cevich mentioned something along the lines of this.

miabbott: I think it would be cleaner if you could stick to one, but I won't hold up merging

mike-nguyen: I think I will change it to all includes for consistency

tests/rpm-ostree/main.yml
mike-nguyen: Added comment

tests/rpm-ostree/main.yml
mike-nguyen: Added comment.

tests/rpm-ostree/main.yml
jlebon: Did you add the reload test, or do you mean to add it in a separate PR?

tests/rpm-ostree/main.yml
```diff
@@ -0,0 +1,565 @@
...
+    - name: Get bootloader entry
+      shell: grep ^initrd /boot/loader/entries/ostree-{{ osname }}-0.conf | sed -e 's,initrd ,/boot/,'
+      register: initrd
+
+    - name: Check initramfs file
+      command: test -n {{ initrd.stdout }}
+
+    - name: Get contents of initrd
+      shell: lsinitrd {{ initrd.stdout }} > lsinitrd.txt
```
jlebon: I'd still change this to use `lsinitrd -f` to make it an even stronger test.

tests/rpm-ostree/main.yml
mike-nguyen: @jlebon I ran into an error running rpm-ostree reload yesterday.  I was going to ask you about it today since you were in training but now I can't reproduce it.  I've added the reload after modifying the origin file.

tests/rpm-ostree/main.yml
mike-nguyen: @jlebon when I run `lsinitrd $initrd -f /my/file` it produces no file.  Am I running the command right?

```
[root@fedora ~]# ll
total 16
-rw-------. 1 root root 5538 Apr 19 13:13 anaconda-ks.cfg
-rw-------. 1 root root 5196 Apr 19 13:13 original-ks.cfg
[root@fedora ~]# lsinitrd /boot/ostree/fedora-atomic-58da926b4642c6974a8b357b074f3ad3cf168163fb28c42cf3a010eda0ab4b84/initramfs-4.11.0-0.rc6.git0.1.fc26.x86_64.img -f ./myfile.txt
[root@fedora ~]# ll
total 16
-rw-------. 1 root root 5538 Apr 19 13:13 anaconda-ks.cfg
-rw-------. 1 root root 5196 Apr 19 13:13 original-ks.cfg
[root@fedora ~]# lsinitrd /boot/ostree/fedora-atomic-58da926b4642c6974a8b357b074f3ad3cf168163fb28c42cf3a010eda0ab4b84/initramfs-4.11.0-0.rc6.git0.1.fc26.x86_64.img 
Image: /boot/ostree/fedora-atomic-58da926b4642c6974a8b357b074f3ad3cf168163fb28c42cf3a010eda0ab4b84/initramfs-4.11.0-0.rc6.git0.1.fc26.x86_64.img: 49M
========================================================================
Early CPIO image
========================================================================
drwxr-xr-x   3 root     root            0 Jan  1  1970 .
-rw-r--r--   1 root     root            2 Jan  1  1970 early_cpio
drwxr-xr-x   3 root     root            0 Jan  1  1970 kernel
drwxr-xr-x   3 root     root            0 Jan  1  1970 kernel/x86
drwxr-xr-x   2 root     root            0 Jan  1  1970 kernel/x86/microcode
-rw-r--r--   1 root     root        24070 Jan  1  1970 kernel/x86/microcode/AuthenticAMD.bin
========================================================================
Version: dracut-044-178.fc26

Arguments: --reproducible --gzip -v --add 'ostree' --tmpdir '/tmp' -f --no-hostonly --add 'iscsi' --kver '4.11.0-0.rc6.git0.1.fc26.x86_64' --reproducible --gzip -v --add 'ostree' --tmpdir '/tmp' -f -I '/etc/rpmostree-initramfs-testing-file' --kver '4.11.0-0.rc6.git0.1.fc26.x86_64'
..snip...
```

jlebon: Heh, sorry, I should have written it as `-f $path_to_my_file`. I.e. the same path you included earlier. E.g.:

```
# echo "barbaz" > /etc/foo.conf
# rpm-ostree initramfs --enable --arg="-I" --arg="/etc/foo.conf"
# lsinitrd /boot/ostree/fedora-atomic-73577e678e475e49881041c3b7c747c9665b95a645cc2aef85e6678c62a5ce35/initramfs-4.11.3-200.fc25.x86_64.img -f /etc/foo.conf
barbaz
```

tests/rpm-ostree/main.yml
mike-nguyen: I'm not sure if I'm doing something wrong but -f does not seem to do anything.

mike-nguyen:
Much fixups.  I believe I have all the requested changes in now this this last fixup!

jlebon:
Looks good to me! 👍

miabbott:
Let's do this!  :shipit:

PR #165, opened 2017-06-01
Use empty string for extra args if undefined

mike-nguyen:
In #153  support was added for passing extra arguments to the
subscription manager command.  If no value was defined in the
subscription csv for 'extras' the lookup command would return []
which would then be added to the end of the subscription manager
command and fail.  Since the return from the lookup is [], the
jinja2 default filter of empty string did not get applied.

The fix I came up with was to set extra_args to the extras lookup
only if the extra lookup is not [].  Otherwise it will set it to
empty string.  This preserves backwards compatability to the old
subscription csv format for people that are already using the test.

### DIFF
--- a/roles/redhat_subscription/tasks/main.yml
+++ b/roles/redhat_subscription/tasks/main.yml
@@ -36,6 +36,10 @@
     when: subscription_file is not defined
     run_once: true
 
+  - name: Set extra args
+    set_fact:
+      extra_args: "{{ lookup('csvfile', 'extra file={} delimiter=,'.format(subscription_file)) if lookup('csvfile', 'extra file={} delimiter=,'.format(subscription_file)) != [] else '' }}"
+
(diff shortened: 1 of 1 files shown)


miabbott:
The docs for the CSV `lookup` say that you can specify a `default` argument which defines the default return value if the key is not found.

http://docs.ansible.com/ansible/playbooks_lookups.html#the-csv-file-lookup

Could you see if that might work instead of the logic you've proposed here?  It might be cleaner or it might not work at all.

mike-nguyen:
@miabbott the default option worked and is much cleaner.  Thanks for the suggestion.  I tested these scenarios and it seems to work.
1. no extra arg defined in subscription csv
2. extra arg defined in subscription csv

miabbott:
I was going to try to test this out, but just going to get it in to avoid any other blockers.

PR #170, opened 2017-06-05
Expand automated tests on PRs

miabbott:
Since we require that roles/tests work across multiple versions of
Atomic Host, we should start testing across multiple versions of
Atomic Host.

This change enables the running of the `improved-sanity-test` on a
Fedora 25 Atomic Host and CentOS 7 Atomic Host (which is pretty close
to a RHEL Atomic Host).

Granted, we are only running the sanity tests now, but hopefully we
can modify this to run new roles or tests in the future.


jlebon:
This looks sane to me. 👍

cgwalters:
Looks sane to me.

miabbott:
Hey, the checks worked too!  ✔️

mike-nguyen:
Nice!!

Issue #150, opened 2017-05-12
refactor the 'improved-sanity-test' to run on only HEAD


miabbott:
Our testing strategy with the `improved-sanity-test` has always had a known drawback:  if a problem exists in the HEAD-1 commit, the whole test fails and we miss out on testing the bits in the HEAD commit.

We've been able to live with this and workaround it via manual testing, but this is not going to scale.

It's time to re-think the test suite to test only the HEAD commit.  This will require the test suite to synthesize a commit to upgrade to and may not be a 'real-world' upgrade test, but I believe the trade-off is worth it for the coverage we gain.

dustymabe:
does this mean we won't do an 'upgrade' or just that we won't run a ton of tests before the upgrade?

mike-nguyen:
Off the top of my head I think @jlebon has steps for a simulated update so we can check for /etc /var 3-way merges, pkg layering through updates, etc.  Then maybe we can do what you do @dustymabe and upgrade to an older commit for RPM changes.  We're open to ideas because HEAD-1 is pretty confusing if you're looking at statuses and results.

miabbott:
@jlebon showed us an example of this at one point.  It looks something like this (the `sed` line is probably not right):

```
# ostree refs --create local-branch <HEAD checksum>
# sed -i 's/^refspec=.*$/refspec=local-branch/' /path/to/origin/file/<HEAD checksum>.origin
# ostree commit -b local-branch --tree=ref=local-branch --add-metatdata-string version=<version>
# rpm-ostree upgrade
```

miabbott:
![](https://github.trello.services/images/mini-trello-icon.png) [(5) Refactor the [a-h-t] sanity test to only require the HEAD commit to be deployed](https://trello.com/c/i2Dx4nFO/767-5-refactor-the-a-h-t-sanity-test-to-only-require-the-head-commit-to-be-deployed)

dustymabe:
note: for my use case I wanted to be able to run tests on a disk image.. @miabbott already added the ability to do this in https://github.com/projectatomic/atomic-host-tests/commit/f15a61d02c4581be31b6a89bf31b2a64cf305558 with `--tag cloud_image`

mike-nguyen:
#198 Will be merged on June 20th and resolve this issue

miabbott:
Closed via #198

PR #173, opened 2017-06-09
[WIP] Run PAPR on newly added test

mike-nguyen:
Currently PAPR will only run the improved sanity test when there is
a new pull request.  This commit will add a very simple script to
run any new test.


.papr.yml
```diff
@@ -13,8 +13,7 @@ packages:
   - ansible
 
 tests:
-  - ansible-playbook -vi testnode, common/ans_ah_head-1_deploy.yml
-  - ansible-playbook -vi testnode, tests/improved-sanity-test/main.yml
+  - .test_director
```
miabbott: I think this needs to be `./.test_director`

See https://github.com/projectatomic/atomic/blob/master/.redhat-ci.yml#L19 as an example

.test_director
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -xeuo pipefail
+
+NEW_TEST=$((git diff --name-only master | grep 'tests/.*/main.yml') || true)
```
jlebon: This can end up being more than a single test though, right? Should we run this in a loop instead? (Though that would require all tests to make sure to cleanly restore the machine to a clean state after they're done). Or maybe it's safer to only ever target a single test and error out if multiple tests are modified to force multiple PRs to be created. Not sure.

miabbott: There's been some recent work (#142 for example) to make sure our tests leave the system in the original state.  It **should** be safe to run multiple tests in succession.

miabbott:
`./.test_director: line 4: git: command not found`

Looks like we need to add `git` to the list of packages being installed.

miabbott:
```
>>> ./.test_director
++ git diff --name-only master
++ grep 'tests/.*/main.yml'
+ NEW_TEST='tests/fake-test/main.yml
tests/rpm-ostree/main.yml'
+ '[' -z tests/fake-test/main.yml tests/rpm-ostree/main.yml ']'
./.test_director: line 6: [: tests/fake-test/main.yml: binary operator expected
```

Looks like you better handle the multiple changed tests after all.

mike-nguyen:
It looks like the diff also includes the last commit that was merged.  There must be a difference in the way we're pulling in the PR and PAPR.

miabbott:
```
>>> ./.test_director
++ git diff --name-only master
++ grep 'tests/.*/main.yml'
+ NEW_TESTS='tests/fake-test/main.yml
tests/improved-sanity-test/main.yml
tests/rpm-ostree/main.yml'
+ for NEW_TEST in '"${NEW_TESTS[@]}"'
+ '[' -z tests/fake-test/main.yml tests/improved-sanity-test/main.yml tests/rpm-ostree/main.yml ']'
./.test_director: line 7: [: too many arguments
```

I think you may need to use the `PAPR_COMMIT` that @jlebon mentioned [here](https://github.com/projectatomic/atomic-host-tests/issues/171#issuecomment-306823270)

Also, you need a different way to test the `NEW_TESTS` variable since it could be empty, one file, or multiple files.

jlebon:
It depends on your needs, but yeah you probably want `git diff --name-only origin/master..$PAPR_COMMIT` here instead of `master`. Basically, by default PAPR runs its tests on the merge commit of your PR and master. Because your PR branch is 2 commits behind master, the overall diff is including diffs from those commits as well. If you only want the files modified in your PR, then the above should work!

mike-nguyen:
Thanks for the ongoing comments as I'm working my way through this :+1:

.test_director
mike-nguyen: I think we can also end up in a state where a test is deleted.  Since we are not checking the status we could be trying to run a test on something that is no longer there.  Once I work out some of the current issues and have something working I'll add some more logic into it to handle different scenarios.

miabbott:
```
>>> ./.test_director
+ HEAD=0cb112d11292c10d087b395060b498053ec95c60
+ NEW_TESTS=($(git diff --name-only origin/master..$HEAD | grep 'tests/.*/main.yml' || true))
++ git diff --name-only origin/master..0cb112d11292c10d087b395060b498053ec95c60
++ grep 'tests/.*/main.yml'
+ '[' 2 -eq 0 ']'
```

This looks like it is doing the right thing now, but I this PR is based on top of 2edd28a so `git diff` is picking up the rpm-ostree test as part of the PR.

If you rebase the PR onto master (`git pull -r upstream master`), the next time the checks run, they should only be running the fake-test.

mike-nguyen:
The last test actually looked like it picked up improved-sanity-test instead of rpm-ostree--which confused me. I added some more info to debug and also simplified the fake-test that runs.

miabbott:
This looks like it is in a good place.  We can add more logic to test changed roles in another PR.

mike-nguyen:
:+1:

mike-nguyen:
I think I didn't account for the scenario where the test that is changed is the improved-sanity-test which would run the improved-sanity-test without deploying HEAD-1 first.

miabbott:
> I think I didn't account for the scenario where the test that is changed is the improved-sanity-test which would run the improved-sanity-test without deploying HEAD-1 first.

More ammo to work on #150 sooner rather than later

miabbott:
[(5) Configure [a-h-t] PAPR to validate changed roles/tests](https://trello.com/c/2CiYbiFB/766-5-configure-a-h-t-papr-to-validate-changed-roles-tests)

PR #174, opened 2017-06-09
tests: allow users to skip images in docker-build-httpd

miabbott:
I ran into the following bug while testing:

https://bugzilla.redhat.com/show_bug.cgi?id=1460341

...and realized there was no easy way to skip the images that were
failing.  This was frustrating, so I fixed it.

The test can now accept an extra var named `cli_skipped_images` which
will take a string of image names to be skipped by the test.  If the
name of the current image under test is in that string, the test skips
that image.

There is no validation that the string contains valid values or
anything like that, so it is up to the user to be smart when using it.

### DIFF
--- a/tests/docker-build-httpd/main.yml
+++ b/tests/docker-build-httpd/main.yml
@@ -22,7 +22,7 @@
   become: yes
 
   tags:
-    - setup
+    - docker_build_setup
 
   vars_files:
     - vars.yml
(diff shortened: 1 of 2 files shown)


mike-nguyen:
Looks great!

mike-nguyen:
bot, retest this please

miabbott:
The failed `fedora/25/atomic` check might be the Fedora version of this BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1460341

PR #176, opened 2017-06-13
selinux_verify: stop checking backup files

miabbott:
New composes done with `rpm-ostree` will have the backups of the
`passwd` removed, per projectatomic/rpm-ostree#693.  This will cause
the `selinux_verify` role to fail.

Since we are already checking the primary file, consensus says that
it is safe to remove the checks on that particular backup, as well
as the other backups of `group`, `shadow`, and `gshadow`.


cgwalters:
:+1:

dustymabe:
LGTM

Issue #175, opened 2017-06-13
configure 'rpm-install' role to use different RPM


miabbott:
Having to scrape a web page to determine the name of the `fpaste` RPM before installing is cumbersome and introduces more chances of failure.

@dustymabe suggested that we use the `epel-release` RPM which should have a static name based on the major version release.

https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm

dustymabe:
> @dustymabe suggested that we use the epel-release RPM which should have a static name based on the major version release.

correction - I think the url won't change, but the rpm that gets downloaded can have a different NEVR. i.e. today you could download an rpm from that link and it is epel-release-1.0.0 and tomorrow it could be epel-release-2.0.0. Make sense?

miabbott:
FYI, even though the [EPEL](https://fedoraproject.org/wiki/EPEL) page points to dl.fp.o for the `epel-release` RPM, you can use download.fp.o and it works just as well.

https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm

mike-nguyen:
@dustymabe do you know of another rpm that is like epel-release that has a constant URL?

dustymabe:
> @dustymabe do you know of another rpm that is like epel-release that has a constant URL?

I don't. I think that one was special because they wanted to be able to have people pull it directly. This was before `epel-release` was in the base centos repos.

mike-nguyen:
#177 Changes the rpm install for the improved sanity test.  Since there was no other case like epel-release, the admin-unlock test will continue to use the parsed html to find the correct file name from EPEL (this is now split into two roles).

PR #177, opened 2017-06-13
Simplify rpm install verification in improved sanity test

mike-nguyen:
Addresses issue #175.  The previous method of verifying rpm installs
relied scraping HTML to determine the filename for a RPM before it
could install it.  There was also a single point of failure pointing
to dl.fedoraproject.org instead of the mirror site.

This PR simplifies the verification in the improved sanity test by
using the epel-release package in the mirror site which always has
the same URL.


roles/rpm_install_epel/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+# vim: set ft=ansible:
+#
+# This role will install epel7 from download.fedoraproject.org and
+#  installs it using the rpm command.  This hsould be used after
```
miabbott: s/hsould/should/

miabbott:
Instead of a new custom role, I'd rather see the `rpm_install` role reworked to perform an install from a URL.  Then, if we were in the situation where we have to perform some additional work to derive a URL, we could make a separate role to do that.

I'd do something just like this:

```yaml
---                                                                                                                                                                                                                
# vim: set ft=ansible:                                                                                                                                                                                             
#                                                                                                                                                                                                                  
- name: Fail if rpm_url is not defined                                                                                                                                                                             
  fail:                                                                                                                                                                                                            
    msg: "rpm_url must be passed into this role/include"                                                                                                                                                           
  when: rpm_url is undefined                                                                                                                                                                                       
                                                                                                                                                                                                                   
- name: Install RPM from URL                                                                                                                                                                                       
  command:  rpm -i {{ rpm_url }}
```

It would require some more changes to `tests/admin-unlock/main.yml` but there is no reason we can't just use the `epel-release` package again in that test.

WDYT?

mike-nguyen:
Makes too much sense to me.

mike-nguyen:
bot, retest this please

miabbott:
LGTM!

PR #179, opened 2017-06-13
Fix inventory for running improved sanity test

mike-nguyen:
The inventory file was accidentally pointing to a test machine instead
of the testnode.


miabbott:
Bad on me for not seeing that in the last review.  😞

Issue #171, opened 2017-06-06
configure PAPR to run tests in PRs, not just 'improved-sanity-test'


miabbott:
Our PAPR implementation is pretty rudimentary - it will run the `improved-sanity-test` on multiple platforms, which is a start.  But often new PRs are introducing new roles/tests that aren't covered by the `improved-sanity-test`.

I'm thinking we need a helper script to determine if a test has been changed or added, then run that test.

This might  be a little tricky if the PR just modifies a role (although, I guess you could `grep` to see which test used said role), but we can always fallback to just running `improved-sanity-test`

Tagging @jlebon because I feel like he may have helped solve this problem elsewhere.

jlebon:
Right, definitely doable. You can use e.g. `git diff --name-only` to get just the list of filenames modified in the PR, and then apply some heuristics on that. There's an upstream issue open in PAPR to support this natively: https://github.com/jlebon/papr/issues/27.

The closest thing we have right that's "self-aware" is the ostree/rpm-ostree submodule checker: https://github.com/projectatomic/rpm-ostree/blob/8b8bdcc/ci/ci-commitmessage-submodules.sh. One tricky part worth noting is that it uses the PAPR-injected env var `PAPR_COMMIT` to make sure we're not doing our diffs based on the merge commit, but from the PR HEAD itself.

miabbott:
![](https://github.trello.services/images/mini-trello-icon.png) [(5) Configure [a-h-t] PAPR to validate changed roles/tests](https://trello.com/c/2CiYbiFB/766-5-configure-a-h-t-papr-to-validate-changed-roles-tests)

mike-nguyen:
@jlebon  Maybe its a difference in workflow but whenever I do a git diff on origin/master..$PAPR_COMMIT, it picks up files that have been merged since the PR was originally submitted.  The PRs are usually built on top of older commits so they are usually behind HEAD.  Do you usually rebase before pushing a new PR?

From looking quickly at the PAPR code, I think the diff I actually want is between the PR merge and master i.e. git diff origin/master..PAPR_MERGE_COMMIT.  That should give me only the files I have changed on top of master.  WDYT?  My git-fu is novice level so please feel to correct me!

jlebon:
[...]

PR #181, opened 2017-06-14
Use correct git diff to determine tests to run

mike-nguyen:
The two dot notation used originally does not work the same way in
git diff as other git commands.  Git diff expects two endpoints as
opposed to a range.  This was causing the diff to return files that
were not a part of the PR to be included in the automatic tests that
are run when there is a new PR.

### DIFF
--- a/.test_director
+++ b/.test_director
@@ -1,9 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 
-# check commit
-HEAD=${PAPR_COMMIT:-HEAD}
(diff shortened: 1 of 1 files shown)


mike-nguyen:
See #171 for details

miabbott:
That is some git-fu right there.  I'll let the checks complete and then merge it in.

miabbott:
I don't know why I am waiting on the checks...they aren't actually testing this PR.  ¯\_(ツ)_/¯

PR #182, opened 2017-06-15
selinux_verify:  enable support for rawhide

### LINKED ISSUE #180: update 'selinux-verify' to handle Fedora Rawhide
The files/processes that are checked as part of the `selinux-verify` role do not have a case for Fedora Rawhide.  Let's fix that.

miabbott:
The `selinux_verify` role was failing when run against the Rawhide
version of Atomic Host.  Rawhide Atomic Host reports itself as Fedora
27 via Ansible, but has the same files/procs as Fedora 26 (in the
context of this role), so just symlink to the `fedora-26.yml`.

Closes #180

### DIFF
--- a/roles/selinux_verify/vars/fedora-27.yml
+++ b/roles/selinux_verify/vars/fedora-27.yml
@@ -1,0 +1,1 @@
+fedora-26.yml


miabbott:
@dustymabe Easy fix ⬆️

dustymabe:
@miabbott hmm - does that work? just includes the fedora 26 one?

miabbott:
@dustymabe It worked when I tested it against Rawhide

dustymabe:
k, +1

mike-nguyen:
Simple enough.  Merging.

PR #183, opened 2017-06-15
selinux_verify: add support for RHELAH 7.4

miabbott:
In RHELAH 7.4, a few changes made that required an update to the files
being checked for SELinux labels.

This allows to test RHELAH 7.3 and the pre-release version of RHELAH
7.4 at the same time.  When 7.4 is released, we could remove this
specialized file if we decide to.


mike-nguyen:
LGTM.  Worked for me on 7.4.

PR #184, opened 2017-06-15
i-s-t: cleanup any extra deployments at the end

miabbott:
When running the `improved-sanity-test` using `--tags cloud_image` the
system is left with a 'hotfix' deployment in the waiting.  If we want
to use the system again to run additional tests, the nice thing to do
is to cleanup after ourselves.

In the case of normal execution, where the system starts at HEAD-1,
upgrades, than rolls back to HEAD-1, we will end up cleaning the
deployment at HEAD.  I'm of the opinion this isn't a deal-breaker.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -643,6 +643,12 @@
       tags:
         - var_files_present
 
+    # cleanup any extra deployments
+    - role: rpm_ostree_cleanup_all
+      tags:
+        - rpm_ostree_cleanup_all
+        - cloud_image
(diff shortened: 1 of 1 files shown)


miabbott:
In the `cloud_image` case, the remaining deployment still has the `httpd` package layered.  Will have to submit another commit that handles that.

miabbott:
Bah...this still needs some more work.

miabbott:
This is ready for review now

mike-nguyen:
LGTM.  Lets merge this!

PR #185, opened 2017-06-15
atomic_scan_verify:  remove /etc/atomic.d/ workaround

miabbott:
We no longer need to modify `/etc/atomic.d/openscap` because the
OpenSCAP container we use has been fixed.

See [RHBZ#1418464](https://bugzilla.redhat.com/show_bug.cgi?id=1418464)


mike-nguyen:
LGTM

PR #188, opened 2017-06-16
roles: add checks for Docker storage configuration

### LINKED ISSUE #121: add new test to verify default storage driver for docker
@dustymabe suggested we get a test case into the mix for verifying the default storage driver for docker

Based on this BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1436777

F26 was the intended target, but we could construct a test case that validates all the streams

### LINKED ISSUE #166: Verify docker storage persists across reboot
Let's make sure we don't hit [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1457912) again by adding a test for it in a-h-t. 

Can most likely be done as part of the work for https://github.com/projectatomic/atomic-host-tests/issues/121

miabbott:
This role will check for the correct default Docker storage driver and
that any mounts are in place.

Closes #121, Closes #166

### DIFF
--- a/roles/docker_storage_verify/tasks/main.yml
+++ b/roles/docker_storage_verify/tasks/main.yml
@@ -12,16 +12,12 @@
     msg: "The required variables are not defined.  Check files under roles/docker_storage_verify/vars"
   when: driver is undefined
 
-  # Maximum efficiency!  The output from 'docker info' contains a line:
-  #
-  #   Storage Driver: <value>
-  #
-  # So we grab the value, split on the colon + space, and compare the value
-  # to the expected driver.  All in one task!
+  # When all the streams are on docker 1.13, we should start using
+  #  docker info --format  '{{.Driver}}'
 - name: Check Docker storage driver
-  shell: docker info | grep '^Storage Driver'
+  shell: "docker info | grep -oP '(?<=Storage Driver: ).*'"
   register: storage_driver
-  failed_when: "'{{ storage_driver.stdout.split(': ')[1] }}' != '{{ driver }}'"
+  failed_when: "'{{ storage_driver.stdout }}' != '{{ driver }}'"
 
 - name: Check for F25 Docker mount
   command: findmnt /var/lib/docker/devicemapper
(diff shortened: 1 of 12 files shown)


miabbott:
@dustymabe Have a look.  I want to make sure this is sane before I start plumbing it into the sanity test.

Interestingly, only Fedora explicitly displays a device being mounted onto a directory.  RHEL and CentOS don't have that output.

miabbott:
[(3) Add [a-h-t] test case to verify default docker storage driver](https://trello.com/c/7e2MgiLb/711-3-add-a-h-t-test-case-to-verify-default-docker-storage-driver)

roles/docker_storage_verify/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  # Maximum efficiency!  The output from 'docker info' contains a line:
+  #
+  #   Storage Driver: <value>
+  #
+  # So we grab the value, split on the colon + space, and compare the value
+  # to the expected driver.  All in one task!
+- name: Check Docker storage driver
+  shell: docker info | grep '^Storage Driver'
+  register: storage_driver
+  failed_when: "'{{ storage_driver.stdout.split(': ')[1] }}' != '{{ driver }}'"
```
dustymabe: this might be easier:

```
[root@f26-updates-testing ~]# docker info --format '{{json .Driver}}' 
"overlay2"
```

miabbott: 👍 Thanks for the tip!

miabbott: Man, I really wanted to use that but the `--format` flag on `docker info` is only available on 1.13

Once all the streams are on 1.13, I can go back and update this.

I figured a cleaner usage of `grep`, so I'm still going to push an update with that.

dustymabe: i'm that guy only using the latest and greatest of everything and expecting everyone else to just update :) - sorry about that - maybe add a comment to the code to mention updating it when it can be updated

mike-nguyen:
LGTM.  Do you plan on plumbing this role into I-S-T in this PR or another one?

miabbott:
I'm hesitant to add it into i-s-t right now because of #198.  I'll work up another PR that we can land after #198 merges.

mike-nguyen:
I'll merge this in now then.

Issue #156, opened 2017-05-24
consider adding retry logic to operations that require external networks


miabbott:
We've seen failures in the tests due to connection timeouts or abrupt resets of connections when the tests are doing things like `atomic pull` or `docker pull`

We might be able to workaround some of these transient failures by just using the `retries` parameter in the roles.

Example:

```
- name: Pull busybox
  command: atomic pull busybox
  register: pull
  retries: 6
  delay: 30
  until: pull.rc == 0
```

mike-nguyen:
Definitely needed!

miabbott:
![](https://github.trello.services/images/mini-trello-icon.png) [(2) Enhance [a-h-t] network roles to implement retries](https://trello.com/c/AVW7ZB4X/764-2-enhance-a-h-t-network-roles-to-implement-retries)

miabbott:
A bunch of work was done in this space with #190, #193, and #195 

I think it is safe to close.

PR #190, opened 2017-06-21
roles: implement retries and remove version check

miabbott:
We no longer need to check the version of the `atomic` command in
order to determine how to list images with the command, since all the
platforms we test against are at version 1.12+ of `atomic`.

Additionally, I added a retry loop on the `atomic pull` in order to
mitigate any failures due to network problems.

This address a small sliver of #156

### DIFF
--- a/roles/atomic_pull_verify/tasks/main.yml
+++ b/roles/atomic_pull_verify/tasks/main.yml
@@ -9,29 +9,15 @@
 #   - atomic run
 #   - atomic stop
 #
-- name: Get atomic major version
-  shell: atomic --version 2>&1 | cut -d. -f1
-  register: atomic_major
-
-- name: Get atomic minor version
-  shell: atomic --version 2>&1 | cut -d. -f2
-  register: atomic_minor
-
 - name: Pull busybox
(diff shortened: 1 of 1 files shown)


roles/atomic_pull_verify/tasks/main.yml
```diff
@@ -9,29 +9,15 @@
...
-  when: (atomic_minor.stdout|int) < 12 and (atomic_major.stdout|int) == 1
-
-- name: Set atomic images command for 1.12 and newer
-  set_fact:
-    atomic_images: "atomic images list"
-  when: (atomic_minor.stdout|int) >= 12 and (atomic_major.stdout|int) == 1
+  register: apb
+  retries: 5
+  delay: 60
+  until: apb|success
```
mike-nguyen: I love the shortcut for the retries here!

mike-nguyen:
LGTM

miabbott:
[(2) Enhance [a-h-t] network roles to implement retries](https://trello.com/c/AVW7ZB4X/764-2-enhance-a-h-t-network-roles-to-implement-retries)

PR #191, opened 2017-06-22
add support for removing all installed packages

miabbott:
Sometimes you just want all the installed packages removed, so this
introduces support for the `all` keyword in the `rpm_ostree_uninstall`
role.  It is not 100% bulletproof, since it is possible a user could
use the `all` keyword when there is nothing installed and then the
role will probably blow up.  But I'm hoping that we are better than that.

There could be additional work done in the
`rpm_ostree_uninstall_verify` role to support the `all` keyword, too.
But I was unable to arrive a solution that would both verify the
packages were uninstalled and the binaries were removed.

### DIFF
--- a/roles/rpm_ostree_uninstall/tasks/main.yml
+++ b/roles/rpm_ostree_uninstall/tasks/main.yml
@@ -41,31 +41,35 @@
         ros_not_booted: "{{ ros_json['deployments'][0] if ros_json['deployments'][0] is defined else false }}"
       when: ros_json['deployments'][1] is defined and ros_json['deployments'][1]['booted']
 
+    # If the 'packages' key has a value, we create a list of packages to
+    # uninstall, otherwise skip it
     - name: Get all installed packages
       set_fact:
         all_installed: "{{ ros_booted['packages']|join(' ') }}"
+      when: ros_booted['packages']|length > 0
   when: packages == "all"
 
-# If the 'all_installed' variable is defined, we use that otherwise just use
-# the supplied 'packages' variable.  NOTE: 'all_installed' could be an empty
-# string and we don't check for it because I'm making the wrong assumption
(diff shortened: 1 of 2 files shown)


roles/rpm_ostree_uninstall/tasks/main.yml
```diff
@@ -19,12 +20,46 @@
...
+      set_fact:
+        ros_booted: "{{ ros_json['deployments'][1] }}"
+        ros_not_booted: "{{ ros_json['deployments'][0] if ros_json['deployments'][0] is defined else false }}"
+      when: ros_json['deployments'][1] is defined and ros_json['deployments'][1]['booted']
+
+    - name: Get all installed packages
+      set_fact:
+        all_installed: "{{ ros_booted['packages']|join(' ') }}"
+  when: packages == "all"
+
```
mike-nguyen: Maybe we can handle the all case like this:

```
        - name: Get all installed packages
          set_fact:
            all_installed: "{{ ros_booted['packages']|join(' ') }}"
          when: ros_booted['packages']|length > 0
      when: packages == "all"

    - name: Setup 'rm_packages' variable
      set_fact:
        rm_packages: "{{ all_installed if all_installed is defined else packages }}"

    - name: Uninstall {{ rm_packages }}
      command: rpm-ostree uninstall {{ rm_packages }}
      when: not reboot and rm_packages != "all"

    - name: Uninstall {{ rm_packages }} and reboot
      command: rpm-ostree uninstall {{ rm_packages }} -r
      async: 60
      poll: 0
      when: reboot and rm_packages != "all"
      ignore_errors: true

    - include: ../../../common/ans_reboot.yml
      vars:
        skip_shutdown: true # shutdown was already initiated
      when: reboot and rm_packages != "all"
```

mike-nguyen:
LGTM.  Merging.

PR #193, opened 2017-06-22
roles: add retries to network operations

miabbott:
This adds retry logic to roles that rely on external network
operations.  I settled on 5 retries per action with delays of 60s
between each attempt.  This should help mitigate some of the failures
we encounter when the network is flaky.

Additionally, I re-wrote some of the existing retries to follow the
format of:

```
register: <var>
retries: 5
delay: 60
until: <var>|success
```

This makes our retry logic more consistent across roles.

This is another commit to address #156

### DIFF
--- a/roles/atomic_installation_verify/tasks/main.yml
+++ b/roles/atomic_installation_verify/tasks/main.yml
@@ -24,6 +24,10 @@
 
 - name: Install cockpit container with tag
   command: "atomic install {{ cockpit_cname }}:latest"
+  register: aic
+  retries: 5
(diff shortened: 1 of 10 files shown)


miabbott:
[(2) Enhance [a-h-t] network roles to implement retries](https://trello.com/c/AVW7ZB4X/764-2-enhance-a-h-t-network-roles-to-implement-retries)

miabbott:
Super tired of seeing network timeouts in our jobs, so I'm merging this.

deal with it

PR #194, opened 2017-06-27
tests: unsubscribe at end of system-containers test

miabbott:
We should be removing our subscriptions at the end of the
`system-containers` test.


mike-nguyen:
LGTM

PR #197, opened 2017-07-10
Update handling of /etc/sysconfig/docker for docker-latest

### LINKED ISSUE #192: fix 'docker' test to handle additional 'docker-latest' requirements
Seen on CAHC, the version of `docker-latest` (`docker-latest-1.13-28.git6cd0bbe.el7.x86_64`) has additional lines in `/etc/sysconfig/docker` that need to be uncommented for everything to work smoothly.

See below:

```
# docker-latest daemon can be used by starting the docker-latest unitfile.
# To use docker-latest client, uncomment below lines
#DOCKERBINARY=/usr/bin/docker-latest
#DOCKERDBINARY=/usr/bin/dockerd-latest
#DOCKER_CONTAINERD_BINARY=/usr/bin/docker-containerd-latest
#DOCKER_CONTAINERD_SHIM_BINARY=/usr/bin/docker-containerd-shim-latest
```

miabbott:
The more recent versions (1.13, I think) of `docker-latest`
require additional lines to be modified in `/etc/sysconfig/docker` for
everything to play happy together.

Closes #192

### DIFF
--- a/roles/docker_build_httpd/tasks/main.yml
+++ b/roles/docker_build_httpd/tasks/main.yml
@@ -47,6 +47,10 @@
 
 - name: Build httpd image
   command: "docker build -t {{ g_httpd_name }} -f {{ build_dir }}/{{ g_httpd_name }}_Dockerfile {{ build_dir }}"
+  register: db
+  retries: 5
+  delay: 60
+  until: db|success
 
 - name: Get docker images after build
   command: docker images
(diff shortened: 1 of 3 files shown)


miabbott:
The `fedora/25/atomic` test is expected to fail because Fedora Atomic Host doesn't have `docker-latest` installed by default.

miabbott:
>  Does the k8-cluster test need these changes for docker 1.13 also? I noticed the docker_private_registry role modifies /etc/sysconfig/docker also but only to add an insecure registry.

It shouldn't.  These changes are only for enabling `docker-latest` and the `k8-cluster` test doesn't use that.

mike-nguyen:
This looks good to go then.

PR #198, opened 2017-07-11
i-s-t: use synthesized upgrades to test HEAD

mike-nguyen:
Because the current improved sanity test requires the test system
to be at HEAD-1, it made it more complicated to figure out when
issues were introduced.  Consumers of the test results would have
to wait until HEAD+1 to see if patch worked.  This commit aims
to resolve these issues by using a synthesized upgrade provided
by @jlebon.

The synthesized upgrade creates a local branch of the current ostree
then creates a commit with no changes.  At this point running an
upgrade would upgrade to an ostree that does not include
packages that are added/upgraded, but the trade-off is worth it for
better CI tests and less user confusion.  Users can see when their
changes causes breakage and when their patches fix them.

Another issue with using this approach is that the synthesized
upgrades is that the system must be upgraded to the local branch
right away.  Creating the local branch later in the test does not
preserve admin/unlock and layered packages that are added during
the test.


mike-nguyen:
Questions up for discussion:
- Should we make the default running with HEAD-1 and an option to run with synthesized upgrades?
- When the test completes, it leaves the system deployed with the local branch.  Should we rebase to the original refspec at commit at the end of the test as a cleanup?

miabbott:
> Should we make the default running with HEAD-1 and an option to run with synthesized upgrades?

My vote is to remove the `HEAD-1` option completely.  I think moving to this approach of local branch + local commit is both less confusing and more valuable.  I don't think we've ever received feedback saying "Testing from HEAD-1 is a great idea!"  It's always been the opposite.

> When the test completes, it leaves the system deployed with the local branch. Should we rebase to the original refspec at commit at the end of the test as a cleanup?

Yeah, I would vote for this too.  Ideally, all of our tests should return the system to the original state before the tests started.

roles/origin_file_modify/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Get current booted dir
+  command: ostree admin --print-current-dir
+  register: current_dir
+
+- name: Replace current refspec
+  replace:
+    dest: "{{ current_dir.stdout }}.origin"
+    regexp: '^(.*refspec=).*$'
+    replace: '\1{{ refspec }}'
```
cgwalters: Why not use `rpm-ostree rebase` instead?

roles/local_repo_setup/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  when: ros_json['deployments'][0] is defined and
+        ros_json['deployments'][0]['booted']
+
+- name: Set current checksum
+  set_fact:
+    checksum: "{{ ros_json['deployments'][1]['checksum'] }}"
+  when: ros_json['deployments'][1] is defined and
+        ros_json['deployments'][1]['booted']
+
+- name: Create local repo
```
cgwalters: `Create local branch`

roles/local_repo_setup/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  set_fact:
+    ros_json: "{{ ros.stdout|from_json }}"
+
+- name: Set current checksum
+  set_fact:
+    checksum: "{{ ros_json['deployments'][0]['checksum'] }}"
+  when: ros_json['deployments'][0] is defined and
+        ros_json['deployments'][0]['booted']
+
+- name: Set current checksum
```
cgwalters: `Set previous checksum`

roles/local_repo_commit/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+#
+- name: Fail if version, refspec, or branch_name is undefined
+  fail:
+    msg: "version, refspec, or branch_name is undefined"
+  when: version is undefined or
+        branch_name is undefined or
+        refspec is undefined
+
+- name: Create a commit with no changes
+  command: ostree commit -b {{ branch_name }} --tree=ref={{ refspec }} --add-metadata-string version={{ version }}
```
cgwalters: We should probably add an ostree command for this.  I'll keep it in mind.

cgwalters:
Another highlevel approach would be to pair a "last released ref" with each ref we're testing.  For FAH, this currently requires manual intervention because of how we do commit promotion.  But for FAHC, that'd be the upstream fedora stable one for example, rather than the last FAHC commit.

roles/origin_file_modify/tasks/main.yml
mike-nguyen: The strategy was to keep the test as close to the most common use case for a user.  Modifying the origin file allows us to user rpm-ostree upgrade even though rebase would be much easier.

roles/local_repo_setup/tasks/main.yml
mike-nguyen: @cgwalters I think my terminology is slightly off on repo and branch.  I will rename the roles to use branch instead of repo.

roles/local_repo_setup/tasks/main.yml
mike-nguyen: I only want to set the checksum for the booted deployment.  Not booted is a no-op.

cgwalters: Oh, I see!  Hmm, I feel like there's probably a way to do this with jinja filters or something but yep, looks correct as is.

mike-nguyen: Yeah, I would definitely like a cleaner way to do this!  I'll look around to see if I can find anything.

roles/origin_file_modify/tasks/main.yml
miabbott: I figured this was possible, but I couldn't easily find it in any man pages or help...until some Google-fu lead me to this little snippet:

`rpm-ostree is now capable of rebasing to a local branch using the rebase :local-branch syntax`

http://www.projectatomic.org/blog/2017/06/rpm-ostree-v2017.6-released/

Sure, enough after creating the local branch, we can just rebase to that and then do the upgrade:

```
# rpm-ostree status
State: idle
Deployments:
● fedora-atomic:fedora-atomic/25/x86_64/docker-host
                Version: 25.148 (2017-06-25 21:07:13)
                 Commit: f136ff5370c0f77d9c1f4fd3ecb8e1815650d8aaf834e41a5e995a13494407e9
# ostree refs --create local-branch f136ff5370c0f77d9c1f4fd3ecb8e1815650d8aaf834e41a5e995a13494407e9
# rpm-ostree rebase :local-branch
Copying /etc changes: 22 modified, 0 removed, 57 added
Transaction complete; bootconfig swap: yes deployment count change: 1
Run "systemctl reboot" to start a reboot
# rpm-ostree status
State: idle
Deployments:
  local-branch
                Version: 25.148 (2017-06-25 21:07:13)
                 Commit: f136ff5370c0f77d9c1f4fd3ecb8e1815650d8aaf834e41a5e995a13494407e9

● fedora-atomic:fedora-atomic/25/x86_64/docker-host
                Version: 25.148 (2017-06-25 21:07:13)
                 Commit: f136ff5370c0f77d9c1f4fd3ecb8e1815650d8aaf834e41a5e995a13494407e9

... reboot system ...

# ostree commit -b local-branch --tree=ref=local-branch --add-metadata-string version=test1
f1cc261cffc29167bbbed04a3c83ce11c62ebe131dbdd89e11abbc08ad287a8b
# rpm-ostree upgrade
Copying /etc changes: 22 modified, 0 removed, 59 added
Transaction complete; bootconfig swap: no deployment count change: 0
Run "systemctl reboot" to start a reboot
# atomic host status
State: idle
Deployments:
  local-branch
                Version: test1 (2017-07-11 20:42:48)
                 Commit: f1cc261cffc29167bbbed04a3c83ce11c62ebe131dbdd89e11abbc08ad287a8b

● local-branch
                Version: 25.148 (2017-06-25 21:07:13)
                 Commit: f136ff5370c0f77d9c1f4fd3ecb8e1815650d8aaf834e41a5e995a13494407e9
```

mike-nguyen:
> My vote is to remove the HEAD-1 option completely. I think moving to this approach of local branch + local commit is both less confusing and more valuable. I don't think we've ever received feedback saying "Testing from HEAD-1 is a great idea!" It's always been the opposite.

@miabbott I was going back and looking at the tags when it hit me that we can probably remove all the cloud_image tags.  This should be able to run on a cloud_image except in cases where the remote is invalid (RHEL AH?) because we rebase back to the original refspec.

miabbott:
> I was going back and looking at the tags when it hit me that we can probably remove all the cloud_image tags. This should be able to run on a cloud_image except in cases where the remote is invalid (RHEL AH?) because we rebase back to the original refspec.

Yeah, I would actually like to see the entire test worked over to make it cleaner in the case of running from `HEAD`.  Remove extra tags, extra steps, roles, etc.  

The case of RHELAH is tricky...maybe we have to create two local branches, one that represents the 'original' refspec and another to mutate.  Then we can revert back to the 'original' local branch and modify the origin file to make it look like the original content set.

WDYT?

mike-nguyen:
> Yeah, I would actually like to see the entire test worked over to make it cleaner in the case of running from HEAD. Remove extra tags, extra steps, roles, etc.

Maybe this can be done in another PR as this is getting more complicated for the RHELAH case discussed below.

>The case of RHELAH is tricky...maybe we have to create two local branches, one that represents the 'original' refspec and another to mutate. Then we can revert back to the 'original' local branch and modify the origin file to make it look like the original content set.

Definitely tricky.  The cloud image remote isn't valid so we can't rebase back to it later on.   I haven't tried rebasing back to a local branch so I will have to investigate that.  In terms of my current strategy, I was planning on subscribing first to get a valid remote/ref setup and saving those to rebase during the cleanup portion of the test.  That should bring it back to the original package set but pointing to the wrong remote.  How much value do you see trying to get the system back to the invalid remote?

miabbott:
> Maybe this can be done in another PR as this is getting more complicated for the RHELAH case discussed below. I've already removed all cloud_image tags in my pending fixup.

I'm cool with that.  😎 

> How much value do you see trying to get the system back to the invalid remote?

I guess in the RHELAH case, we will probably want to access via a subscription, so I suppose it is not critical.  If we can get the system nearly back to it's original state (i.e single deployment, no installed pkgs, no containers), that would be good enough.  I'd just see how a system behaves when you run another test suite against it after an i-s-t completion.

miabbott:
> Another highlevel approach would be to pair a "last released ref" with each ref we're testing. For FAH, this currently requires manual intervention because of how we do commit promotion. But for FAHC, that'd be the upstream fedora stable one for example, rather than the last FAHC commit.

@cgwalters Could you elaborate on this?  What problems would that solve?

cgwalters:
The synthetic content-less approach is useful, but it'd be a lot more "real" to test upgrades from a commit that has already passed a-h-t.    But testing from a stable commit, we still avoid the issues with `HEAD-1`.  We'd start catching actual upgrade issues, like new code doesn't handle old data in `/var`, etc.

cgwalters:
I have no objections to committing this now, we can consider what I'm talking about a next step.

roles/origin_file_modify/tasks/main.yml
mike-nguyen: Nice find @miabbott.  I just tried this on  RHELAH 7.3.6 but it is running rpm-ostree v2017.5 so we will have to wait until rpm-ostree v2017.6 is in.

miabbott:
These changes are ready to be merged, but we are going to hold off until Thurs July 20, in order for any consumers of the i-s-t to prepare for the change.

miabbott:
[(5) Refactor the [a-h-t] sanity test to only require the HEAD commit to be deployed](https://trello.com/c/i2Dx4nFO/767-5-refactor-the-a-h-t-sanity-test-to-only-require-the-head-commit-to-be-deployed)

PR #199, opened 2017-07-11
Enable Fedora 26 testing via PAPR

miabbott:
Today is the release of Fedora 26, so we should start using it in our
own PAPR tests.


mike-nguyen:
LGTM

PR #201, opened 2017-07-14
remove runc test

mike-nguyen:
This test was not reliable  because of the necessary setup
needed to run the test.  The level of runc coverage needed
is covered by the k8-cluster and openshift test.  The upstream
test [1] is still available for users who want to run it manually.

[1] https://fedorapeople.org/cgit/caiqian/public_git/runctst.git/


miabbott:
I'm OK with this.  Assuming we get any blowback, we can just `git revert`

PR #202, opened 2017-07-14
Update Dockerfiles to use proper FROM field

miabbott:
The Dockerfiles we use in our tests had a myriad of FROM formats, so
this cleans up the state of those.  Now all the FROM lines have the
following:

- fully-qualified domain
- explict tag (:latest or otherwise)

Additionally, the CentOS and Fedora base images are changed to be
pulled from the community registries instead of docker.io


cgwalters:
LGTM

mike-nguyen:
Looks good!

PR #206, opened 2017-07-14
add support for /etc/containers/registries.conf

mike-nguyen:
Supported is needed for an additional  scenario now for adding
registries and insecure registries.  The current way to add
registries is in /etc/sysconfig/docker.  The new way is to add
registries to /etc/containers/registries.conf which uses yaml
format.  If /etc/containers/registries.conf exists, it the
registry will be added in the file, otherwise it will be added
in /etc/sysconfig/docker.

### DIFF
--- a/roles/docker_private_registry/tasks/main.yml
+++ b/roles/docker_private_registry/tasks/main.yml
@@ -9,17 +9,32 @@
     when: ansible_docker0.ipv4.address is not defined
     run_once: true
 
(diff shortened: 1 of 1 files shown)


miabbott:
Yah, looks good  👍

miabbott:
FWIW, the failures in the CentOS are expected because of known issues in those distros being used.

PR #207, opened 2017-07-14
Remove broken distros from papr.yml

miabbott:
The `centos/7/atomic` and `centos/7/atomic/smoketested` distros are
having issues that are affecting our PAPR tests.  I've removed them
temporarily until they are stable again.

Related issues:

CentOS/sig-atomic-buildscripts#282

CentOS/sig-atomic-buildscripts#283

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -1,13 +1,11 @@
 cluster:
   hosts:
     - name: testnode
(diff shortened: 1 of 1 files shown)


mike-nguyen:
Makes sense

PR #211, opened 2017-07-20
redhat_subscription: remove unconfigured-state after registering

miabbott:
When provisioning RHELAH hosts in our interal Jenkins jobs, we end up
with an origin file that has an `unconfigured-state` present.  And
when the system is registered via `subscription-manager`, that line is
not removed as part of the process.

This changes the `redhat_subscription` role to remove the
`unconfigured-state` line in the current origin file.  If the line is
already missing, this is no-op.


mike-nguyen:
LGTM.

PR #213, opened 2017-07-26
Add support for checking the journal for strings

### LINKED ISSUE #186: added AVC denial check to 'improved-sanity-tests'
We should put some checks in the `improved-sanity-test` to look for AVC denials in the journal.

Offhand, I'd say do a check before and after every boot/reboot.  This should let us catch any denials that happen during boot or any that were silently ignored before the system reboots.

cc: @dustymabe

miabbott:
This introduces two roles that can be used for checking the journal of
a host for particular strings.

`journal_search`:  A role that will search the journal for a specific
                   string supplied to the role.  Optionally can specify
                   that the role fails when the string is found.

`journal_fatal_msgs`:  A 'meta' role that calls out to `journal_search`
                       to look for certain phrases that are considered
                       fatal to the overall test.

The `journal_fatal_msgs` role is used in the `improved-sanity-test` to
inspect the state of the journal before and after the host is
rebooted during the test duration.

Closes #186

### DIFF
--- a/roles/journal_fatal_msgs/meta/main.yml
+++ b/roles/journal_fatal_msgs/meta/main.yml
@@ -1,0 +1,16 @@
+---
+# vim: set ft=ansible:
+#
+# Use this role to specify any strings or phrases that should be considered
+# fatal if they appear in the journal of a host under test.
+#
+dependencies:
+  # check for SELinux denials
+  - role: journal_search
+    search_string: denied
+    fail_if_found: true
+
+  # check for kernel Oops
+  - role: journal_search
+    search_string: Oops
+    fail_if_found: true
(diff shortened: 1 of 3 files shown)


miabbott:
[(3) Add checks for AVC denials as part of the [a-h-t] sanity test](https://trello.com/c/vq5MqpYq/782-3-add-checks-for-avc-denials-as-part-of-the-a-h-t-sanity-test)

mike-nguyen:
LGTM

dustymabe:
Does this just do a blanket check for denials? - Freaking awesome guys thanks so much! :tada:

miabbott:
@dustymabe Pretty much.  It runs before/after every reboot during the sanity test.  It will currently check for SELinux denials, kernel oops, and unmapped SELinux contexts.

dustymabe:
> @dustymabe Pretty much. It runs before/after every reboot during the sanity test. It will currently check for SELinux denials, kernel oops, and unmapped SELinux contexts.

I suspect we'll be opening a lot of bugs, but that is a good thing. I just hope we can get the selinux team to fix them quick enough.

PR #218, opened 2017-08-02
Minor improvements to checking the journal for fatal messages

miabbott:
This updates the SELinux denial string that is sought to be more precise
to denials related to SELinux.

Additionally, I added more tags to the `journal_fatal_msgs` roles used
in the `improved-sanity-test` to allow better selection of which roles
should or should not run.

### DIFF
--- a/roles/journal_fatal_msgs/meta/main.yml
+++ b/roles/journal_fatal_msgs/meta/main.yml
@@ -7,7 +7,7 @@
 dependencies:
   # check for SELinux denials
   - role: journal_search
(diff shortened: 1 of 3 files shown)


mike-nguyen:
Looks good.  Merging!

miabbott:
[(3) Add checks for AVC denials as part of the [a-h-t] sanity test](https://trello.com/c/vq5MqpYq/782-3-add-checks-for-avc-denials-as-part-of-the-a-h-t-sanity-test)

PR #219, opened 2017-08-03
roles:  fix b_d_s_f to work with local branches

miabbott:
If the `booted_deployment_set_fact` role is used after the host has
been upgraded/rebased to a local branch, it will currently fail
because the refspec has no `remote` portion.

This fixes the role to work with a local branch by checking for the
`remote:refspec` syntax (just the presence of the `:` actually) and
then setting up the facts appropriately.

### DIFF
--- a/roles/booted_deployment_set_fact/tasks/main.yml
+++ b/roles/booted_deployment_set_fact/tasks/main.yml
@@ -25,6 +25,8 @@
 
 - name: Save refspec and checksum
   set_fact:
-     saved_checksum: "{{ ros_booted['checksum'] }}"
-     saved_refspec: "{{ ros_booted['origin'].split(':')[1] }}"
-     saved_remote_name: "{{ ros_booted['origin'].split(':')[0] }}"
+    saved_checksum: "{{ ros_booted['checksum'] }}"
+    # If the host is on a local branch, we have to skip splitting the 'origin'
+    # because there is no remote portion
+    saved_refspec: "{{ ros_booted['origin'].split(':')[1] if ':' in ros_booted['origin'] else ros_booted['origin'] }}"
+    saved_remote_name: "{{ ros_booted['origin'].split(':')[0] if ':' in ros_booted['origin'] else '' }}"


miabbott:
@mike-nguyen In the local branch case, I opted to set the value of `saved_remote_name` to an empty string, but could also be convinced to set it to the the `origin` value.

mike-nguyen:
@miabbott I think empty string for `saved_remote_name` is fine.  We can change it to the `origin` value if we find a need for it.  Looks good to me overall.

miabbott:
@mike-nguyen ffffffffuuuuuuuuuuu

```
    "attempts": 5, 
    "changed": true, 
    "cmd": [
        "rpm-ostree", 
        "rebase", 
        ":local-branch", 
        "beec4d8ba6051173794c568acd972e6545f0fb45364df4ecc09fc9366076a0a3"
    ], 
    "delta": "0:00:00.031282", 
    "end": "2017-08-08 14:54:26.688112", 
    "failed": true, 
    "rc": 1, 
    "start": "2017-08-08 14:54:26.656830"
}
```

From this:

```
    "attempts": 5, 
    "changed": true, 
    "cmd": [
        "rpm-ostree", 
        "rebase", 
        ":local-branch", 
        "beec4d8ba6051173794c568acd972e6545f0fb45364df4ecc09fc9366076a0a3"
    ], 
    "delta": "0:00:00.031282", 
    "end": "2017-08-08 14:54:26.688112", 
    "failed": true, 
    "rc": 1, 
    "start": "2017-08-08 14:54:26.656830"
}
```


Guess we do need the remote name...

miabbott:
This is a bit more complicated because of what I am trying to do in #220...

It looks like the `saved_refspec` and `saved_remote_name` values used to be set to the `remote:refspec` value that the host started with at the beginning of the `improved-sanity-test`.  But when I start using the `booted_deployment_set_fact` role in `rpm_signature_verify`, the values are over-written by the `local-branch` value.

I'll need to figure out how to save those starting values.

PR #220, opened 2017-08-03
roles: add ability to check RPM signatures

miabbott:
The three platforms which we support with these tests have periodic
official releases where the RPMs are all signed by a well known GPG
key.  When testing these releases/streams, we want to be able to
automtically verify that the RPMs have been signed with the correct
GPG key.

This role adds support for verifying that the RPMs are correctly
signed.  In its current implementation, we only support checking for a
single GPG signature, which is how the RPM content has been
historically signed.  In other words, none of the platforms currently
sign their RPMs with multiple keys for an official release.

This role has also been added to the `improved-sanity-test`.


miabbott:
Note: this is dependent on #219

miabbott:
[(3) Add a check for RPM signatures to [a-h-t]](https://trello.com/c/EdGMRP99/807-3-add-a-check-for-rpm-signatures-to-a-h-t)

miabbott:
F25 is failing with an SELinux denial and F26 is failing because it needs the fix in #219

mike-nguyen:
bot, retest this please

miabbott:
This has uncovered some additional scenarios that need to be accounted for in other roles.  Using that `WIP` label for now.

miabbott:
This should be good to go once #225 is merged.  Then this should be re-run through the CI.

mike-nguyen:
bot, retest this please

mike-nguyen:
LGTM.  Worked nicely on RHELAH also.

Issue #217, opened 2017-08-02
fix callback plugin to suppress stdout when 'no_log: true' is used


miabbott:
In https://github.com/projectatomic/atomic-host-tests/blob/master/roles/journal_search/tasks/main.yml#L24, we use the `no_log: true` directive to stop the entirety of the journal from being displayed during the test execution.

But in our automated jobs, it appears that this is not happening because the callback plugin is separately printing out `STDOUT` itself.

miabbott:
Closed via #222

PR #222, opened 2017-08-08
hide stdout and stderr when no_log is true

mike-nguyen:
Resolves issue #217 

Ansible has a no_log option that will hide stdout and stderr when
set to true.  The callback plugin was not respecting this option
when formatting the ansible output.  This commit checks the result
_ansible_no_log value before adding it to the output.


miabbott:
LGTM!

mike-nguyen:
:+1:

PR #223, opened 2017-08-08
roles: set 'saved_remote_name' to 'origin' value

miabbott:
When the host is using a local-branch as its deployment, there is no
remote name by default.  This may cause problems when attempting to
rebase the system.  To work around this, we use the `origin` value as
the remote.

### DIFF
--- a/roles/booted_deployment_set_fact/tasks/main.yml
+++ b/roles/booted_deployment_set_fact/tasks/main.yml
@@ -29,4 +29,4 @@
     # If the host is on a local branch, we have to skip splitting the 'origin'
     # because there is no remote portion
     saved_refspec: "{{ ros_booted['origin'].split(':')[1] if ':' in ros_booted['origin'] else ros_booted['origin'] }}"
-    saved_remote_name: "{{ ros_booted['origin'].split(':')[0] if ':' in ros_booted['origin'] else '' }}"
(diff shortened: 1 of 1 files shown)


miabbott:
TBH, I'm not sure this is required, because I uncovered a whole host of other problems while working on this.  But I think it makes sense.

mike-nguyen:
Which scenarios would be affected by having a blank origin value?  In the local-branch scenario, rebasing to local-branch:local-branch would fail because no remote named local-branch is defined.

miabbott:
When working on a new test, I found that the final rebase in `improved-sanity-test` was failing because it was trying to do:

`# rpm-ostree reabse :local-branch`

This was because I was using the `booted_deployment_set_fact` role as a `meta` dependency in a new role and it was overwriting the `saved_*` facts that was used to get back to the original deployment.

I addressed getting back to the original deployment in #225 and I thought this would complement it.

This change might not be necessary.  TBH, I've had a lot of changes in flight lately and this could just be a side-effect of a wonky situation I found myself in.

miabbott:
We could punt on this change and alter `rpm_ostree_rebase` to detect if there is a non-empty string for the `remote` name and then do `rpm-ostree rebase <branch>` and omit the remote

mike-nguyen:
Makes sense

PR #225, opened 2017-08-09
Save original deployment info for later use

miabbott:
It's possible that through the course of testing that the original
deployment information is lost (more acccurately, not easily
available) via upgrades/rebases.  In order to return the host under
test to its original state, we need a way to retain the original
deployment information to be recalled later.

This problem is solved by writing out the deployment information
(`checksum`, `refspec`, and `remote`) to an INI-style file in the
local playbook directory.  This file can then be read in by another
role that will setup variables to be used by the `rpm_ostree_rebase`
role.

The solution is admittedly a bit specific to this problem/variables,
but I think could be modified in the future to be a bit more
versatile.


miabbott:
I'm overwriting the `saved_*` facts with this change, but it might be better to save off new facts that we can use for the 'rpm_ostree_reabse` role (and default to the `saved_*` facts in that case)

miabbott:
This should probably be merged after #223

roles/original_refspec_set_fact/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+# vim: set ft=ansible:
+#
+# This role configures some facts based on the presence of an INI file that
+# holds some key=value pairs related to the original deployment on a host
+# under test.  This role is designed to be used after the role
+# `booted_deployment_set_fact` has been run.
+#
+# (I will probably regret making such a specific role...)
+#
```
mike-nguyen: Can we template the variables as a yaml file and pull it in with include_vars?  Then we can have a generic role that pulls in vars from a specified file.  TBH, I haven't worked much with templating so I don't know if this will work.

miabbott: I'll look into this approach; sounds promising!

miabbott:
Pushed a new commit that uses a YAML file to load in the original deployment information dynamically ⬆️ 

Great suggestion @mike-nguyen!

miabbott:
The F25 failure is due to SELinux denial on the host;  seems unrelated to this PR.

roles/booted_deployment_set_fact/tasks/main.yml
```diff
@@ -3,30 +3,54 @@
...
+  when: bdsf_ros_json['deployments'][0] is defined and bdsf_ros_json['deployments'][0]['booted']
 
 - name: Set ros variable if deployment 1 is booted
   set_fact:
-    ros_booted: "{{ ros_json['deployments'][1] }}"
-    ros_not_booted: "{{ ros_json['deployments'][0] if ros_json['deployments'][0] is defined else false }}"
-  when: ros_json['deployments'][1] is defined and ros_json['deployments'][1]['booted']
+    bdsf_ros_booted: "{{ bdsf_ros_json['deployments'][1] }}"
+    bsdf_ros_not_booted: "{{ bdsf_ros_json['deployments'][0] if bdsf_ros_json['deployments'][0] is defined else false }}"
+  when: bdsf_ros_json['deployments'][1] is defined and bdsf_ros_json['deployments'][1]['booted']
```
mike-nguyen: Sorry one more suggestion.  I filed #230 to update the rpm_ostree_status and rpm_ostree status_verify to use json_query.  I think that would cleanup this.  Since we only care about the booted deployment, we can do

```
- name: Set booted deployment
  set_fact:
    bdsf_ros_booted: "{{ item }}"
  with_items: "{{ bdsf_ros_json | json_query(query) }}"
  vars:
    query: "[?booted=='true']"
```
Untested so the query might need some work since I don't know if it respects boolean values or reads them as strings.  I believe everything below should work.

miabbott:
`json_query` FTW!  Pushed a fixup :arrow_up: 

In addition to the use of `json_query`, I also opted to go with an empty string for `bdsf_saved_remote_name` in the case of the local branch, since that is 1) more accurate and 2) works better with the `rpm_ostree_rebase` role now that #227 is merged.

miabbott:
Welp, now we need another package for PAPR before we can start using the `json_query` filter

See #231

mike-nguyen:
bot, re-test this please

mike-nguyen:
bot, retest this please

mike-nguyen:
Looks good!  Great work Mr Miabbott!

miabbott:
:fireworks: :champagne: :+1:

PR #228, opened 2017-08-14
.papr.yml: Add CentOS AH back to PAPR coverage

miabbott:
We were having issues deploying `HEAD-1` of the CentOS AH stream,
because the commit history of that stream had an unsigned commit in
it.  Since we no longer need to deploiy `HEAD-1` for the purposes of
running the `improved-sanity-test`, we can add the broken stream back
in.

Note:  I believe the `centos-atomic-host/7/x86_64/devel/smoketested`
branch still has issues, so I am leaving this out for now.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -26,3 +26,14 @@ cluster:
 
 context: fedora/26/atomic
 
+---
+inherit: true
+
+cluster:
+  hosts:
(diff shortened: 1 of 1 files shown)


mike-nguyen:
LGTM

PR #229, opened 2017-08-15
callback_plugins: init '_ansible_no_log' in plugin

### LINKED ISSUE #226: ansible.plugins.callback.default.CallbackModule object at 0x7f45e46a0b10>): '_ansible_no_log'
Been seeing this error while working on additional roles:

```
[WARNING]: Failure using method (v2_runner_on_unreachable) in callback plugin (<ansible.plugins.callback.default.CallbackModule object at 0x7f45e46a0b10>): '_ansible_no_log'
```

...and it appears to be hiding an underlying error in the Ansible role/task.

I edited `callback_plugins/default.py` and removed the check for `_ansible_no_log` and the real error in my code was revealed:

```
The conditional check '{{ remote_name|length }} == 0' failed. The error was: error while evaluating conditional ({{ remote_name|length }} == 0): 'remote_name' is undefined

The error appears to have been in '/home/miabbott/workspaces/projectatomic/atomic-host-tests/roles/rpm_ostree_rebase/tasks/main.yml': line 47, column 5, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

- block:
  - name: Rebase to commit on refspec
    ^ here

```

In this case, the `remote_name` variable was not defined.

Not sure if this is something we can work around in the plugin?

miabbott:
The `_ansible_no_log` variable does not appear to be defined by
default by Ansible.  This was causing the callback plugin to suppress
any error messages about other undefined variables in a task, as
described in #226.

This change causes the variable to be initialized to `False` if the
variable is found to be undefined.  Doing so allows the other
undefined errors to bubble up.

Closes #226

### DIFF
--- a/callback_plugins/default.py
+++ b/callback_plugins/default.py
@@ -55,6 +55,9 @@ class CallbackModule(DEFAULT_MODULE.CallbackModule):  # pylint: disable=too-few-
         '''Return the text to output for a result.'''
         result['_ansible_verbose_always'] = True
 
+        if '_ansible_no_log' not in result.keys():
+            result['_ansible_no_log'] = False
+
         save = {}
         for key in ['stdout', 'stdout_lines', 'stderr', 'stderr_lines', 'msg']:
             if key in result:


mike-nguyen:
LGTM

PR #231, opened 2017-08-16
papr: install 'python2-jmespath'

miabbott:
If we want to use the `json_query` filter in our roles (and we do), we
need the `jmespath` library installed.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -10,6 +10,7 @@ context: fedora/25/atomic
 packages:
   - ansible
   - git
+  - python2-jmespath
 
 tests:
   - ./.test_director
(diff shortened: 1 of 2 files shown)


mike-nguyen:
Should we update the requirements.txt file also?

mike-nguyen:
https://github.com/ansible/ansible/issues/24319

miabbott:
In my virtualenv, `jmespath` appears to be pulled in automagically via `pip install ansible`.  But since we are installing Ansible via `dnf` in PAPR, that dependency does not exist.

So I think our `requirements.txt` is fine for now.

miabbott:
Nope, I lied.  I have my virtualenv overloaded with some OpenStack modules too, so probably a good idea to add to `requirements.txt`

mike-nguyen:
LGTM

PR #232, opened 2017-08-17
Delete container before uninstall

mike-nguyen:
There is a new requirement that does not allow uninstall if a
container is using the image.  See [1] for details.  This includes
exited containers.  This was causing a failure in this test so
now the exited container is deleted before uninstallation.

[1] https://github.com/projectatomic/atomic/pull/1050


miabbott:
Makes sense.  Let's see if we can get the bot to retest it.

miabbott:
bot, retest this please

mike-nguyen:
bot, u da bestest

PR #233, opened 2017-08-17
roles: init the 'rsv_rpm_sig_supported' fact

miabbott:
It turns out that you can't conditionally set a fact in Ansible and
provide a default at the same time.  So we initialize the fact first
to a default variable and then apply the conditial to set it.`


miabbott:
F25 failure is due to an SELinux denial unrelated to the PR

mike-nguyen:
I just tried it on CAHC and this check works.

PR #234, opened 2017-08-17
papr:  add CAHC to the list of hosts

miabbott:
In #220, I tried to get fancy with some Ansible conditionals and
default filters, but failed to test all the possible paths, resulting
in false failures.  One of the missing paths was testing with CAHC, so
I think it would be useful to include that in the automated PAPR
checks.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -38,3 +38,15 @@ cluster:
     image: registry.fedoraproject.org/fedora:25
 
 context: centos/7/atomic
+
+---
+inherit: true
+
+cluster:
+  hosts:
+    - name: testnode
+      distro: centos/7/atomic/continuous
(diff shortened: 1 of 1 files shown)


miabbott:
The CAHC failure is because #233 wasn't merged in before this change.

mike-nguyen:
The CAHC stream moves a bit fast for us to know whether the test is broken or the stream is broken.  I do see value in having it here for us to look at the logs but not enough to gate on it.

miabbott:
> The CAHC stream moves a bit fast for us to know whether the test is broken or the stream is broken. I do see value in having it here for us to look at the logs but not enough to gate on it.

Agreed...a little more work for us, but still valuable.

PR #235, opened 2017-08-17
tests: drop 'rpm_signature_verify' after upgrade

miabbott:
At the beginning of the `improved-sanity-test`, we run the
`rpm_signature_verify` check to ensure that the supported streams are
checked to have the correct RPM signatures.  We skip this check on
certain streams which are known not to always have RPMs signed by a
single key.

Because the upgrade is performed against a local branch created from
the original commit, the content of the local branch is already well
known and has been checked for RPM signatures.  Thus we can skip the
RPM signature check for the second time.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -375,11 +375,6 @@
         - journal_fatal_msgs_second_boot
         - journal_fatal_msgs
 
-    # Check RPM signatures again
-    - role: rpm_signature_verify
-      tags:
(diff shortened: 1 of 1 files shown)


miabbott:
@mike-nguyen This is the easy way out.  Alternatively, we may be able to improve `rpm_signature_verify` to handle the post-upgrade case, but I wonder about the feasibility.

dustymabe:
LGTM

mike-nguyen:
LGTM

PR #236, opened 2017-08-24
Update fedora vars files now that 27 has branched.

dustymabe:
Now that fedora 27 has branched from rawhide we need to add
a fedora 28 profile. Also let's update the generic fedora symlink
to point to fedora-28.


mike-nguyen:
LGTM

PR #240, opened 2017-08-28
improved-sanity-test not improved-sanity-tests

guillaumevincent:
improved-sanity-tests folder doesn't exist
replace by improved-sanity-test


miabbott:
Good catch!  Thanks for the fix!

PR #244, opened 2017-09-07
roles: refactor 'rpm_ostree_status*' to use json_query filter

### LINKED ISSUE #230: update rpm ostree status and verify role to use json_query
The current rpm ostree status and verify roles check assumes only two deployments while setting the booted deployment fact.  With livefs, there can be three deployments so its probably a good idea to update these roles to use json_query.

It will probably look something like:
```

- name: Get rpm-ostree status output
  command: rpm-ostree status --json
  register: ros_status

- name: Import JSON
  set_fact:
    ros_json: "{{ ros_status | from_json }}"

- name: Get matching list entry
  set_fact:
    booted_deployment: "{{ item }}"
  with_items: "{{ ros_json | json_query(query) }}"
  vars:
    query: "[?booted=='true']"
```
Then the individual properties can be grabbed using booted_deployment['checksum'], booted_deployment['origin'], etc.  This needs to be compatible with the tests that use the role already unless all the tests are updated also.

miabbott:
This attempts to streamline the `rpm_ostree_status` role by employing
the `json_query` filter.  We can query the `deployments` from the
output of `rpm-ostree status --json` for the correct deployment which
is booted or not booted.  In the case of a single deployment, we
default to setting the `ros_not_booted` to false.

The `rpm_ostree_status_verify` has been converted to use the
json_query filter as well.  In this role, the `rpm_ostree_status` is
symlinked into the directory, so it can be included directly.  This
was done because we want to have that role run every time the 'verify'
role is called.  (I tried to use the meta dependencies, but it would
only run once per playbook if the 'verify' role was called multiple
times)

I also snuck it a little fix to the `admin-unlock` test that I noticed
while checking to see if the all the tests still worked.

Closes #230

### DIFF
--- a/roles/rpm_ostree_status/tasks/main.yml
+++ b/roles/rpm_ostree_status/tasks/main.yml
@@ -11,16 +11,19 @@
 
 - name: Convert to JSON
   set_fact:
-    ros_json: "{{ ros.stdout|from_json }}"
+    ros_json: "{{ ros.stdout | from_json }}"
 
-- name: Set ros variable if deployment 0 is booted
+- name: Set ros_num_deployments
   set_fact:
-    ros_booted: "{{ ros_json['deployments'][0] }}"
-    ros_not_booted: "{{ ros_json['deployments'][1] if ros_json['deployments'][1] is defined else false }}"
-  when: ros_json['deployments'][0] is defined and ros_json['deployments'][0]['booted']
+    ros_num_deployments: "{{ ros_json|json_query('length(deployments)') }}"
 
-- name: Set ros variable if deployment 1 is booted
+- name: Set ros_booted variable for multi-deployments
   set_fact:
-    ros_booted: "{{ ros_json['deployments'][1] }}"
-    ros_not_booted: "{{ ros_json['deployments'][0] if ros_json['deployments'][0] is defined else false }}"
-  when: ros_json['deployments'][1] is defined and ros_json['deployments'][1]['booted']
+    ros_booted: "{{ item }}"
+  with_items: "{{ ros_json | json_query('deployments[?booted]') }}"
+
+- name: Set ros_not_booted variable
+  set_fact:
+    ros_not_booted: "{{ item | default(false) }}"
+  with_items: "{{ ros_json | json_query('deployments[?!booted]') }}"
+
--- a/roles/reboot/meta/main.yml
+++ b/roles/reboot/meta/main.yml
@@ -1,0 +1,2 @@
+---
+allow_duplicates: true
--- a/roles/rpm_ostree_status/meta/main.yml
+++ b/roles/rpm_ostree_status/meta/main.yml
@@ -1,0 +1,2 @@
+---
+allow_duplicates: true
--- a/roles/rpm_ostree_status_verify/meta/main.yml
+++ b/roles/rpm_ostree_status_verify/meta/main.yml
@@ -1,0 +1,5 @@
+---
+allow_duplicates: true
+
+dependencies:
+  - role: rpm_ostree_status
--- a/roles/rpm_ostree_status_verify/tasks/main.yml
+++ b/roles/rpm_ostree_status_verify/tasks/main.yml
@@ -17,8 +17,8 @@
 # We use 'include:' here instead of a 'meta' dependency because the meta
 # dependency will only be run once if this role is used multiple times in
 # a playbook.
-- name: Pull in the 'rpm_ostree_status' role
-  include: rpm_ostree_status.yml
+#- name: Pull in the 'rpm_ostree_status' role
+#  include: rpm_ostree_status.yml
 
 - name: Verify number of deployments
   fail:
--- a/roles/rpm_ostree_status_verify/tasks/main.yml
+++ b/roles/rpm_ostree_status_verify/tasks/main.yml
@@ -7,44 +7,28 @@
 #   deployment (int) REQUIRED - index of deployment to verify
 #   expected (dict) REQUIRED - a dict of expected key/value pairs
 #   num_deployments (int) OPTIONAL - the number of deployments in output
-
+#
 - name: Fail if deployment or expected is undefined
   fail:
     msg: "deployment or expected is undefined"
   when: expected is undefined or
         deployment is undefined
 
-- name: Get rpm-ostree status output
-  command: rpm-ostree status --json
-  register: ros_status
-
-- name: Set json output
-  set_fact:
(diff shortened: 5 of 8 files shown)


miabbott:
Hmm....CAHC is failing due to projectatomic/rpm-ostree#976.  The other failures are going to require additional debug.

miabbott:
Manual testing didn't give me any real fixes (and different failures), but I did improve some messaging and conditionals.  Fingers crossed!  🤞

miabbott:
Part of the `rpm-ostree` test looks like it may have been broken with this change.  Using `WIP` for now...

roles/rpm_ostree_status/tasks/main.yml
```diff
@@ -11,16 +11,27 @@
...
+  - name: Set ros_booted variable for multi-deployments
+    set_fact:
+      ros_booted: "{{ item }}"
+    with_items: "{{ ros_json | json_query('deployments[?booted]') }}"
+
+  - name: Set ros_not_booted variable
+    set_fact:
+      ros_not_booted: "{{ item | default(false) }}"
+    with_items: "{{ ros_json | json_query('deployments[?!booted]') }}"
+  when: ros_num_deployments > 1
```
jlebon: I'm surprised this block doesn't work equally well when `ros_num_booted == 1`. Is that not the case?

miabbott: Yeah, your logic is better than mine!  😁

miabbott: Oh, I remember why I did it like this.  When the user does `rpm-ostree ex livefs`, we end up with three deployments and I was trying to determine the 'not-booted' one.  But we actually have two 'not-booted', so this is wrong anyways.  😬

miabbott:
>Try using allow_duplicates: true in the meta/main.yml file to have the dependency run everyone the main role is run.

I did give that a try, but no joy.

mike-nguyen:
> Try using allow_duplicates: true in the meta/main.yml file to have the dependency run everyone the main role is run.
>> I did give that a try, but no joy.

It worked for me although I haven't pushed it to a repo yet ¯\_(ツ)_/¯

Did you have the `allow_duplicates: true` in roles/rpm_ostree_status/meta/main.yml and roles/rpm_ostree_status_verify/meta/main.yml?

miabbott:
> It worked for me although I haven't pushed it to a repo yet ¯_(ツ)_/¯
> 
> Did you have the allow_duplicates: true in roles/rpm_ostree_status/meta/main.yml and roles/rpm_ostree_status_verify/meta/main.yml?

Huh.  I'll revisit this.  I may have only had it in one place.

miabbott:
@mike-nguyen if you have something working, push to your fork and share the link

mike-nguyen:
Check out https://github.com/mike-nguyen/atomic-host-tests/tree/atomic_tests  The last two plays don't work but I pushed it into my repo so you can check it out.

miabbott:
I played with `allow_duplicates: true` in both roles, but got failures when running the `rpm-ostree` test suite.

I think it is because we are using `include:` to explicitly load in the role file.  I experimented with `include_role:` and that looks like it works better.  I'll make another commit that uses that approach in the near future.

miabbott:
The latest CAHC failure looks like we caught a new commit during the test; we expected the host to be at the original commit it started with, but because we used `rpm-ostree upgrade` (expecting to go from `STARTCOMMIT-1` to `STARTCOMMIT`) we got the latest commit from the remote.

It looks like the F25 test failed for a similar reason.

And F26 failed because of https://pagure.io/atomic-wg/issue/330

---

Using `rpm-ostree upgrade` here:

https://github.com/projectatomic/atomic-host-tests/blob/master/tests/rpm-ostree/main.yml#L474

...appears to be the culprit of the failures.  This could continue to bite us in the future because the cloud image we start from in the OpenStack infra may not always be up to date with `HEAD` of the various streams we test against.

I think we could mitigate this if we switch to `rpm-ostree deploy STARTCOMMIT` rather than the upgrade.  I understand why it would be preferable to test the upgrade path, but the deploy code path effectively does the same thing.  And at this point of the test, we are just unwinding what has been previously done, I think it is a good compromise.

@mike-nguyen WDYT?

mike-nguyen:
@miabbott At that point we are just cleaning up so doing an rpm-otsree deploy STARTCOMMIT is fine by me!

miabbott:
I cleaned up the commits down to two because the amount of fixups were getting out of control.

miabbott:
Apparently I suck at `git` since some changes got lost during my rebase

¯\_(ツ)_/¯

miabbott:
Well, the CAHC check passed, which was giving me fits before.

- F26 is a known failure:  https://github.com/projectatomic/rpm-ostree/issues/976
- F25 can't find a commit...might be an infra issue; gonna wait the weekend and retry it

miabbott:
The issue with F25 is related to the starting image that PAPR uses when running the checks.

In the internal infra, the image for F25 came from https://kojipkgs.fedoraproject.org/compose/twoweek/ rather than the "official" F25 image.  The "twoweek" image uses a commit + version that does not match the commit + version that exists in the F25 branch on the remote.

Thus, when the test itself tries to fetch the parent of the booted F25 commit, it cannot find the booted commit in the remote and falls over.

miabbott:
@mike-nguyen I think this PR is ready to get reviewed again.  The PAPR failures are known and explainable.

It's definitely grown in scope (unfortunately), so I could try to split out some of the commits into separate PRs if you like.

Removing the WIP label now.

miabbott:
[(3) use json_query in the a-h-t tests when dealing with 'rpm-ostree status'](https://trello.com/c/CKUxT8Xq/845-3-use-jsonquery-in-the-a-h-t-tests-when-dealing-with-rpm-ostree-status)

miabbott:
[(3) use json_query in the a-h-t tests when dealing with 'rpm-ostree status'](https://trello.com/c/CKUxT8Xq/845-3-use-jsonquery-in-the-a-h-t-tests-when-dealing-with-rpm-ostree-status)

mike-nguyen:
bot, retest this please

miabbott:
The F26 failure should go away with the next 2 week release that is due this week.  Once that lands, I'll re-run that specific context.

F25 is dead; long live F25...so let's hand wave that failure away

miabbott:
Rebased onto master and squashed some fixups too ⬆️

mike-nguyen:
This looks good to go.

PR #245, opened 2017-09-12
selinux_verify:  update files/procs for latest 7.4 content

miabbott:
The CAHC stream started picking up RPMs from the CentOS 7.4 release,
which resulted in a few changes to the list of files and processes we
want to check.

### DIFF
--- a/roles/selinux_verify/vars/centosdev-7.yml
+++ b/roles/selinux_verify/vars/centosdev-7.yml
@@ -1,25 +1,22 @@
 ---
 distro_files:
   - { key: '/usr/bin/docker', value: 'system_u:object_r:container_runtime_exec_t:s0' }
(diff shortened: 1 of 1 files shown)


miabbott:
The F25 failure is due to AVC denials unrelated to the PR.  Currently investigating CAHC failure...

miabbott:
The CAHC check is running against an older version of the continuous stream which hasn't been switched over to the updated package set.  When the infra has a more up-to-date version of CAHC available, I'll re-run the check.

miabbott:
Re-ran CAHC check and it passed!  I think this is ready to be merged.

miabbott:
We're falling behind on tracking CAHC, so I'm going to merge this in.

YOLO

PR #247, opened 2017-09-20
roles:  use new regexp method to append kube options

miabbott:
The previous approach to adding required options to the `KUBELET_ARGS`
did not take into account that there could be default options passed
in as args.  This changes uses the `lineinfile` module and its regexp
capabilities to match the `KUBELET_ARGS` line in the file and append
the options we want to the existing line.

Additionally, the `--config` flag for the kubelet was marked as
'Deprecated' way back in Aug 2016[0] and then completely removed in Jan
2017[1].  The replacement option is `--pod-manifest-path`.

[0] https://github.com/kubernetes/kubernetes/commit/e43ccdbf2c0b69bfa19108a4d82894962c108016
[1] https://github.com/kubernetes/kubernetes/commit/bec6635ccc05b1f02c9971ec9bf529faac0b9ca1

### DIFF
--- a/roles/kubernetes_setup/tasks/main.yml
+++ b/roles/kubernetes_setup/tasks/main.yml
@@ -26,11 +26,31 @@
       - controller-mgr-pod.json
       - scheduler-pod.json
 
+  # Thank you based Google gods
+  # https://groups.google.com/d/msg/ansible-project/JvHfchsgRaU/Vw_CzBbvadgJ
+  #
+  # Shoddy explanation of regexp:
+  #   - open of capture group 1: matched by (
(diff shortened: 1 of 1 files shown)


miabbott:
[(3) fix kube test/roles in a-h-t to work with 1.5 and 1.6](https://trello.com/c/S7vI0qNJ/849-3-fix-kube-test-roles-in-a-h-t-to-work-with-15-and-16)

mike-nguyen:
bot, retest this please

mike-nguyen:
LGTM.  Also tested on Fedora 26 and looked great.

PR #250, opened 2017-09-22
Move subscription-manager auto-attach parameter in subscription file

### LINKED ISSUE #249: Missing org_id and activationkey lookup for redhat subscription
Hello,

I got an issue with actual tests. I running `improved-sanity-test` against a `RHEL-Atomic-7.4` virtual machine. My virtual machine is already activated.

So I got one question:

 * Why are we trying to run the `subscription-manager` against a machine with an active subscription?
   If we want to force the subscription to be sure the `subscription-manager` works, could we move this to another test?

If this is not possible, could we fix `roles/redhat_subscription/tasks/main.yml` to lookup for `org_id` and `activationkey` also.

I need to be able to activate my Red Hat subscription with `org_id` and `activationkey` only.

Thanks

guillaumevincent:
subscription_data.csv.sample got an extra parameter.
We can use this extra parameter to subscribe with activationkey for example.

    extra,--activationkey='...' --org='...'

If we do so, roles/redhat_subscription/tasks/main.yml:46 raise an error:

    "Error: Activation keys cannot be used with --auto-attach."

Move --auto-attach in subscription file sample.
Remove --force because extra param is already set

This PR fixes #249

### DIFF
--- a/roles/redhat_subscription/files/subscription_data.csv.sample
+++ b/roles/redhat_subscription/files/subscription_data.csv.sample
@@ -2,4 +2,4 @@ baseurl,https://cdn.redhat.com
 serverurl,subscription.rhn.redhat.com:443/subscription
 username,username@example.com
 password,yourPa$sW0rd
-extra,--force
+extra,--auto-attach --force
--- a/roles/redhat_subscription/tasks/main.yml
+++ b/roles/redhat_subscription/tasks/main.yml
@@ -45,7 +45,7 @@
 
   - name: Register with subscription-manager
     command: >
-      subscription-manager register --auto-attach --force
+      subscription-manager register
       --baseurl="{{ lookup('csvfile', 'baseurl file={} delimiter=,'.format(subscription_file)) }}"
       --serverurl="{{ lookup('csvfile', 'serverurl file={} delimiter=,'.format(subscription_file)) }}"
       --username="$USERNAME"


miabbott:
@guillaumevincent If we remove the `--auto-attach` flag from the extra parameters, shouldn't we also have a `subscription-manager attach` command elsewhere in the role?

guillaumevincent:
@miabbott my PR added the `--auto-attach` flag into the extra parameters. So I don't understand the question. It's the opposite.

I just allow users to remove it if like me they want to subscribe with activation key.

guillaumevincent:
It's not visible in the pull request but extra parameters are called at the end of the command: https://github.com/projectatomic/atomic-host-tests/blob/ce4ed930a5709d522bf06a2710579ef51eeca0c5/roles/redhat_subscription/tasks/main.yml#L53

miabbott:
@guillaumevincent My mistake; I don't know how I got confused like that.  ¯\_(ツ)_/¯

I'll have the bot run tests against this PR.

miabbott:
bot, add author to whitelist

guillaumevincent:
It looks like 3 builds failed because of https://github.com/projectatomic/rpm-ostree/issues/233

`Register with subscription-manager` is ignored on Centos and Fedora. Only used on RHEL

miabbott:
@guillaumevincent Unfortunately, our method of testing PRs can also uncover issues on the platforms we are testing, as we are seeing here:

- The `centos/7/atomic` test failed because when doing `rpm-ostree install httpd`, it pulls in `centos-logos` which wants to install `/boot/grub/splash.xpm.gz`.  This is not currently supported in the version of `rpm-ostree` included in CentOS AH.  See https://github.com/projectatomic/rpm-ostree/issues/853 which shows that there was a fix made only 18 days ago.

- The `fedora/25/atomic` failure is due to an case of a file with the `unlabeled_t` SELinux context.  Additionally, we are phasing out testing PRs against Fedora 25, so this checkpoint will be gone in the future.

- The `fedora/26/atomic` failure appears to be another SELinux issue, where the test is claiming that an AVC denial was found in the journal.  I don't think is is related to this PR, so we can waive this failure.

I'll put it on my todo list to checkout this PR locally and run some of my own tests just to be certain, but I'm very confident this PR will be merged in the next day or two.

guillaumevincent:
Thanks for the update @miabbott

miabbott:
@guillaumevincent I ran this PR locally against RHEL, Fedora 26, and CentOS AH Continuous without trouble.  Thanks for the patience!

PR #251, opened 2017-09-22
roles: update SELinux checks on CentOS AH

miabbott:
CentOS AH was recently updated to match the RHEL release, so we have
to adjust our SELinux checks.


cgwalters:
LGTM

PR #252, opened 2017-09-25
README: remove Fedora 25 results

miabbott:
We've stopped testing Fedora 25 on our internal infra, so we can
remove these results from the dashboard.


jlebon:
Maybe let's drop it from `.papr.yml` as well?

miabbott:
> Maybe let's drop it from .papr.yml as well?

Doh!  Will do

jlebon:
LGTM! 👍

mike-nguyen:
LGTM also!

PR #255, opened 2017-09-26
Fix 404 not found links in doc

guillaumevincent:


guillaumevincent:
We probably can add a `RHEL subscription` subsection in the doc and link every tests/*/README.md to the subsection instead of duplicating Prerequisites everywhere

miabbott:
> We probably can add a RHEL subscription subsection in the doc and link every tests/*/README.md to the subsection instead of duplicating Prerequisites everywhere

Good idea; I opened #256 to track this (among other doc improvements)

miabbott:
LGTM!

PR #261, opened 2017-10-02
tests: install a different package on CentOS

### LINKED ISSUE #254: temporarily find an alternate pkg to install on CentOS AH
We're seeing failures when running the `improved-sanity-test` on CentOS AH when the test tries to do an `rpm-ostree install httpd`

The latest version of CentOS AH had the `centos-logos` package dropped out of its content set, but `httpd` requires it to be installed.  See https://github.com/CentOS/sig-atomic-buildscripts/pull/297#issuecomment-329200716 for a description of what happened with that package.

We should temporarily find another suitable package to install, until https://github.com/projectatomic/rpm-ostree/issues/853 makes it into CentOS AH.

miabbott:
This PR includes two changes:

- improved support for checking if a package has been installed or uninstalled

The `rpm_ostree_install*` and `rpm_ostree_uninstall*` roles have been enhanced so that they can be run multiple times in the same playbook.  Additionally, the `*verify` roles can now check for a binary that does not match the package name.  For example, check for `ntpd` when `ntp` is installed.

- conditional install of the `ntp` package on CentOS

The version of `rpm-ostree` in CentOS doesn't support layering files on `/boot` and this is causing the install of `httpd` to fail because of the `centos-logos` dependency that gets pulled in.  We'll temporarily test package layering using the `ntp` package on CentOS until CentOS gets a version of `rpm-ostree` that supports sticking files in `/boot`.

Closes #254


tests/improved-sanity-test/main.yml
```diff
@@ -481,9 +481,17 @@
...
       packages: httpd
       reboot: false
+      when: ansible_distribution != 'CentOS'
       tags:
         - rpm_ostree_install
 
+    - role: rpm_ostree_install
+      packages: ntp
+      reboot: false
+      when: ansible_distribution == 'CentOS'
```
cgwalters: This seems OK, but isn't it more usual Ansible style to derive variables in `vars.yml` and then reference those?

miabbott: Yeah, that sounds right.  I'll fix that.

miabbott:
[(3) fix package layering test in i-s-t on CentOS AH](https://trello.com/c/rTEa74mn/853-3-fix-package-layering-test-in-i-s-t-on-centos-ah)

miabbott:
I found an issue when running the `pkg-layering` test suite using this PR....investigating the root cause.

miabbott:
The `fedora/26/atomic` issue should hopefully get resolved in the next 2 week release due this week; we can retry that context once the release lands.

miabbott:
I think what is happening on `centos/7/atomic` is that the `pkg-layering` test has a section where it installs a non-root package (`httpd`) and uses the reboot flag.  Because the reboot flag is used, an asynchronous Ansible command is used and the result is never inspected.  Since installing `httpd` on CentOS AH will fail (the whole reason for this now massive PR), the reboot is never initiated.  This causes the playbook to fail because it is waiting for the host under test to go down.

My opinion is that this particular part of the `pkg-layering` playbook should be left intact and the `centos/7/atomic` failure is ignored.

mike-nguyen:
> I think what is happening on centos/7/atomic is that the pkg-layering test has a section where it installs a non-root package (httpd) and uses the reboot flag. Because the reboot flag is used, an asynchronous Ansible command is used and the result is never inspected. Since installing httpd on CentOS AH will fail (the whole reason for this now massive PR), the reboot is never initiated. This causes the playbook to fail because it is waiting for the host under test to go down.
>
>My opinion is that this particular part of the pkg-layering playbook should be left intact and the centos/7/atomic failure is ignored.

I agree since its a valid failure on the test.  The error in the log throws you off but that's the nature of commands that automatically initiate reboots in Ansible.

Overall this LGTM.  Great work on cleaning it up and adding in additional functionality.

miabbott:
Rebased on master and squashed a fixup in the process ⬆️

mike-nguyen:
bot, retest this please

PR #263, opened 2017-10-05
roles/bdsf: prepend 'ansible_host' to archive file

### LINKED ISSUE #258: 'booted_deployment_set_fact' does not play nice when run in parallel
There is an option in the `booted_deployment_set_fact` role where it will drop a YAML file in the playbook directory.  Unfortunately, the file is generically named `bdsf_archive.yml` which means it can overwritten if you run the same playbook multiple times in parallel.

Probably could work around this by just appending the host IP to the file name.

miabbott:
If you ran a long test that used the `booted_deployment_set_fact` role
across multiple hosts in parallel, any archive file that was created
by the test would be clobbered by the last execution that completed.
This creates a problem when you are trying to read that file and
expecting it to have certain values.

To avoid this, the archive file is created with the value of
`ansible_host` prepended to the file name.  This keeps the file unique
to the host under test and should avoid any clobbering that may have
happened.

Closes #258

### DIFF
--- a/roles/booted_deployment_set_fact/tasks/main.yml
+++ b/roles/booted_deployment_set_fact/tasks/main.yml
@@ -39,11 +39,11 @@
 - block:
   - name: Remove existing archive file
     local_action:
-      file path="{{ playbook_dir }}/bdsf_archive.yml" state=absent
+      file path="{{ playbook_dir }}/{{ ansible_host }}_bdsf_archive.yml" state=absent
     become: false
 
   - name: Archive refspec and checksum
     local_action:
-      template src=bdsf_archive.j2 dest="{{ playbook_dir }}/bdsf_archive.yml"
+      template src=bdsf_archive.j2 dest="{{ playbook_dir }}/{{ ansible_host }}_bdsf_archive.yml"
     become: false
   when: bdsf_archive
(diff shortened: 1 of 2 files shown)


miabbott:
Labeling this as WIP since I didn't fully test the impact it will have on `improved-sanity-test`, so failures may occur in the automated checks

mike-nguyen:
Makes sense and code looks pretty straight forward.

miabbott:
> Makes sense and code looks pretty straight forward.

Unfortunately, the value of `ansible_host` changes because it is originally used as a `local_action` and the next time the value is of the remote host.

I'm leaning towards generating a temp dir and file name that can be used.

mike-nguyen:
Can we set a fact at the beginning of booted_deployment_set_fact to the value of ansible_host then use that value when we save it down in the local action?

miabbott:
> Can we set a fact at the beginning of booted_deployment_set_fact to the value of ansible_host then use that value when we save it down in the local action?

Yeah, that's similar to what we do with the `reboot` role

miabbott:
I didn't go with the custom file/directory approach, but this still works.  WIP removed.

mike-nguyen:
LGTM!

PR #265, opened 2017-10-06
README: show F27 testing results

miabbott:
The internal Jenkins instance will soon start producing the results of
testing the F27 testing branch.  We should display these results in
our janky dashboard once they are live.


miabbott:
The Jenkins job isn't live yet, so hold off on any merges until we have real results to display.

miabbott:
Job is live and results are in.  Merging this.

PR #268, opened 2017-10-25
system-containers: disable etcd/flanneld if rpms are installed

miabbott:
On certain flavors of Atomic Host, we still ship the `etcd` and
`flannel` RPMs on the host.  This causes a minor conflict when we
install those services from system containers.  Notably, the existing
services can be running and attempts to `runc exec` into the
containers will fail because they haven't been started yet.

So we check for the presence of the RPMs and proactively disable/stop
the services before trying to use the system containers.

### DIFF
--- a/tests/system-containers/main.yml
+++ b/tests/system-containers/main.yml
@@ -391,6 +391,30 @@
     - vars.yml
 
   tasks:
+    - name: Check for etcd rpm
+      command: rpm -q etcd
+      register: rpm_etcd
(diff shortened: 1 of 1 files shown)


mike-nguyen:
I initially thought we could just change the name option in the atomic install --system command but if etcd/flannel is already running, you cannot start the system container service.  So LGTM :+1:

cgwalters:
LGTM.

PR #269, opened 2017-10-25
tree-wide: correct use of Jinja2 templating in when: statements

miabbott:
I've observed warnings thrown by Ansible about using Jinja2 templating
in `when:` and `failed_when:` statements like this:

```
   when: "'{{ foo }}' in bar.stdout"
```

The correct way to write this is like so:

```
   when: foo in bar.stdout
```

This change corrects (most) of the instances of this to the correct
format.

### DIFF
--- a/docs/howto_contribute.md
+++ b/docs/howto_contribute.md
@@ -79,7 +79,7 @@ operation is successful.
 - name: Verify Docker image was built
   command: docker images
   register: docker_images
-  failed_when: "'{{ image_name }}' not in docker_images.stdout"
+  failed_when: image_name not in docker_images.stdout
 ```
 
 This role is very flexible as it allows users the ability to define the
--- a/roles/atomic_images_delete_all/tasks/main.yml
+++ b/roles/atomic_images_delete_all/tasks/main.yml
@@ -11,4 +11,4 @@
 - name: Verify no images exist
   command: atomic images list -q
   register: ail
-  failed_when: "{{ ail.stdout_lines | length }} > 0"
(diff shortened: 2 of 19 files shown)


miabbott:
I ran almost all of our tests against a F26 Atomic Host, so this **shouldn't** break anything

miabbott:
The failure in the `centos/7/atomic` context is unrelated to this PR.  When performing the install of `httpd` manually, I observed the following:

```
# rpm-ostree install httpd                                                
Checking out tree 33b4f04... done                   
Enabled rpm-md repositories: base updates extras epel                                                    
rpm-md repo 'base' (cached); generated: 2017-09-05 21:51:13                                              
rpm-md repo 'updates' (cached); generated: 2017-10-26 21:19:13                                           
rpm-md repo 'extras' (cached); generated: 2017-10-26 21:08:15                                            
rpm-md repo 'epel' (cached); generated: 2017-10-26 13:31:31                                              

Importing metadata [===============================================================================] 100%
Resolving dependencies... done                      
error: Unpacking centos-logos-70.0.6-3.el7.centos.noarch: Unsupported path: /boot/grub/splash.xpm.gz; See https://github.com/projectatomic/rpm-ostree/issues/233
```

miabbott:
Also, running the remaining portion of the 'pkg-layering' playbook against a `centos/7/atomic` host was successful.

mike-nguyen:
These changes look pretty straight forward.  I think the PAPR runs are enough to confirm that the logic is and the jinja2 formatting is correct.

PR #270, opened 2017-10-27
selinux_verify: log any files/procs with incorrect SELinux labels

### LINKED ISSUE #260: print out matching files/directories when doing SELinux label checks
In the `selinux_verify` role, we just output:

`The file /var had an SELinux label of 'unlabeled_t'`

But we should be able to print out any and all files/subdirs that match by just saving off the stdout and printing it in the `fail:` statement

See also #239 and #253

miabbott:
Previously, the `selinux_verify` role would fail with a generic
message about finding a file or process that had an incorrect SELinux
label.  This caused additional work for anyone doing triage of the
tests and made folks unhappy.

This change alters how the role reports any instances of incorrect
SELinux labels.  The files/directories/processes are saved into a list
which can then be read out later as part of a failure message.

This should result in last failure message to include the relevant
details of the files/processes and negate the need to dig through the
logs further during triage.

Closes #260

### DIFF
--- a/roles/selinux_verify/tasks/main.yml
+++ b/roles/selinux_verify/tasks/main.yml
@@ -51,7 +51,9 @@
 
 - name: Append failed common file labels to the list
   set_fact:
-    sv_failed_file_labels: "{{ sv_failed_file_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
+    sv_failed_file_labels: |
+      {{ sv_failed_file_labels|default([]) +
+      [ item.item.key + ': ' + item.stdout + ' (expected: ' + item.item.value + ')' ] }}
   when: item.item.value not in item.stdout
   with_items: "{{ common_selinux_file_labels.results }}"
 
@@ -62,7 +64,9 @@
 
 - name: Append failed distro file labels to the list
   set_fact:
-    sv_failed_file_labels: "{{ sv_failed_file_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
+    sv_failed_file_labels: |
+      {{ sv_failed_file_labels|default([]) +
+      [ item.item.key + ': ' + item.stdout + ' (expected: ' + item.item.value + ')' ] }}
   when: item.item.value not in item.stdout
   with_items: "{{ distro_selinux_file_labels.results }}"
 
@@ -93,7 +97,9 @@
 
 - name: Append incorrect common SELinux process labels
   set_fact:
-    sv_failed_proc_labels: "{{ sv_failed_proc_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
+    sv_failed_proc_labels: |
+      {{ sv_failed_proc_labels|default([]) +
+      [ item.item.key + ': ' + item.stdout + ' (expected: ' + item.item.value + ')' ] }}
   when: item.item.value not in item.stdout
   with_items: "{{ common_selinux_proc_labels.results }}"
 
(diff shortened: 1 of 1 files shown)


miabbott:
Rebased ⬆️

roles/selinux_verify/tasks/main.yml
```diff
@@ -57,12 +60,29 @@
...
   with_items: "{{ distro_files }}"
 
-- name: Verify distribution specific SELinux file labels are correct
-  fail:
-    msg: "The file {{ item.item.key }} had an incorrect label of {{ item.stdout }} (expected {{ item.item.value }})"
+- name: Append failed distro file labels to the list
+  set_fact:
+    sv_failed_file_labels: "{{ sv_failed_file_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
   when: item.item.value not in item.stdout
   with_items: "{{ distro_selinux_file_labels.results }}"
```
mike-nguyen: We lost the expected value which is also also useful in the triage process but I think it can still be added.  Maybe something like:
```
set_fact:
  sv_failed_file_labels: "{{ sv_failed_file_labels|default([]) + [ item.item.key + ': ' + item.stdout + '(expected:' + item.item.value + ')' ] }}"
```

roles/selinux_verify/tasks/main.yml
```diff
@@ -71,9 +91,9 @@
...
   with_items: "{{ common_procs }}"
 
-- name: Verify common SELinux process labels are correct
-  fail:
-    msg: "The process {{ item.item.key }} had an incorrect label of {{ item.stdout }} (expected {{ item.item.value }})"
+- name: Append incorrect common SELinux process labels
+  set_fact:
+    sv_failed_proc_labels: "{{ sv_failed_proc_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
   when: item.item.value not in item.stdout
   with_items: "{{ common_selinux_proc_labels.results }}"
```
mike-nguyen: Same comment as above.

roles/selinux_verify/tasks/main.yml
```diff
@@ -82,30 +102,63 @@
...
   with_items: "{{ distro_procs }}"
 
-- name: Verify distribution specific SELinux process labels are correct
-  fail:
-    msg: "The process {{ item.item.key }} had an incorrect label of {{ item.stdout }} (expected {{ item.item.value }})"
+- name: Append incorrect distro SELinux process labels
+  set_fact:
+    sv_failed_proc_labels: "{{ sv_failed_proc_labels|default([]) + [ item.item.key + ': ' + item.stdout ] }}"
   when: item.item.value not in item.stdout
   with_items: "{{ distro_selinux_proc_labels.results }}"
```
mike-nguyen: Same comment as above

roles/selinux_verify/tasks/main.yml
miabbott: Heh.  I was wondering if someone was going to mention that.  Should be easy to add that in.

miabbott:
Addressed the suggestion from @mike-nguyen in the fixup ⬆️ 

FWIW, the failed output looks like this:

```
TASK [selinux_verify : Fail if any SELinux file labels were incorrect] *********                                                                                                                                   
fatal: [f26]: FAILED! => {                                                                                                                                                                                         
    "changed": false,                                                                                                                                                                                              
    "failed": true                                                                                                                                                                                                 
}                                                                                                                                                                                                                  
                                                                                                                                                                                                                   
MSG:
---
The following files had incorrect SELinux labels:

/usr/bin/docker: system_u:object_r:container_runtime_exec_t:s0 (expected: system_u:object_r:container_runtime_foo_t:s0)
/usr/bin/container-storage-setup: system_u:object_r:bin_t:s0 (expected: system_u:object_r:foo_t:s0)

---
```

Much improved, IMO!

mike-nguyen:
Looks like you can copy and paste it right into a BZ!

miabbott:
Retrying the `centos/7/atomic` tests...

mike-nguyen:
Looks good.  Ready to go.

PR #271, opened 2017-10-30
i-s-t: use diff RPM for 'rpm_install' on Fedora

miabbott:
Previously, we were getting away with using `epel-release` as a
generic RPM duing the `rpm_install` role in the i-s-t.  But the latest
version of the RPM now has a `Conflicts: fedora-release` and thus we
can't install it on Fedora-based AH systems anymore.

This change conditionalizes the RPM we use during the `rpm_install`
role.  On Fedora, I chose the current stable version of `ltrace` and
stuck with `epel-release` for everything else.

### DIFF
--- a/tests/improved-sanity-test/vars/common.yml
+++ b/tests/improved-sanity-test/vars/common.yml
@@ -8,7 +8,9 @@ g_dir2: 'qe2'
 g_hostname: 'myhostname'
 g_pkg: "wget"
 g_rpm_url: 'https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm'
+g_rpm_url_fedora: 'https://kojipkgs.fedoraproject.org//packages/ltrace/0.7.91/22.fc26/x86_64/ltrace-0.7.91-22.fc26.x86_64.rpm'
 g_rpm_name: 'epel-release'
+g_rpm_name_fedora: 'ltrace'
 g_seboolean: "virt_use_nfs"
 # These variables are used by the 'semanage_fcontext_mod' and
 # 'semanage_fcontext_verify' roles.  The values were pulled from
--- a/tests/improved-sanity-test/vars/common.yml
+++ b/tests/improved-sanity-test/vars/common.yml
@@ -8,7 +8,7 @@ g_dir2: 'qe2'
 g_hostname: 'myhostname'
 g_pkg: "wget"
 g_rpm_url: 'https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm'
-g_rpm_url_fedora: 'https://kojipkgs.fedoraproject.org//packages/ltrace/0.7.91/22.fc26/x86_64/ltrace-0.7.91-22.fc26.x86_64.rpm'
+g_rpm_url_fedora: 'https://kojipkgs.fedoraproject.org/packages/ltrace/0.7.91/22.fc26/x86_64/ltrace-0.7.91-22.fc26.x86_64.rpm'
 g_rpm_name: 'epel-release'
 g_rpm_name_fedora: 'ltrace'
 g_seboolean: "virt_use_nfs"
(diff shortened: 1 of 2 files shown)


miabbott:
@mike-nguyen This fixes our CI on Fedora, so please have a look and merge if appropriate

tests/improved-sanity-test/vars/common.yml
```diff
@@ -8,7 +8,9 @@ g_dir2: 'qe2'
 g_hostname: 'myhostname'
 g_pkg: "wget"
 g_rpm_url: 'https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm'
+g_rpm_url_fedora: 'https://kojipkgs.fedoraproject.org//packages/ltrace/0.7.91/22.fc26/x86_64/ltrace-0.7.91-22.fc26.x86_64.rpm'
```
mike-nguyen: nit, only one slash between fedoraproject.org and packages in the url

Will this url always be there or will it be bumped by new versions at some point?

cgwalters: Since it's in a "gold" release it'll be around quite a long time...you can still get 
https://koji.fedoraproject.org/koji/buildinfo?buildID=1088
from 11 years ago.

cgwalters: (A trickier question is whether the f26 version of ltrace will still install on e.g. F29AH)

miabbott: > (A trickier question is whether the f26 version of ltrace will still install on e.g. F29AH)

Yeah, I was avoiding that question for now.  I welcome any suggestions for `noarch` packages that we could feed into this role.

mike-nguyen: I've been searching but haven't found anything yet.  I am okay with merging this until we find a better package to use for Fedora, CentOS, and RHEL AH.

cgwalters: I looked briefly but couldn't find one.  Should be fine, if it breaks we can deal with it later.

miabbott: Errr....`noarch` would be nice, but probably wouldn't solve the version issue.  I guess we are still on the hunt for a 'latest' URL that would deliver a the latest version of an RPM like `epel-release` does.

cgwalters: Those URLs are what's in the rpm-md returned by e.g.
```
[fedora]
name=Fedora $releasever - $basearch
failovermethod=priority
#baseurl=http://download.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
```

And really this is what e.g. `rpm-ostree install ltrace` is doing.  I think though this role falls into a gap where we're using `rpm` directly on top of `ostree admin unlock`  (We should definitely have that as an rpm-ostree feature..., it's sort of there with livefs except there it's persistent)

Anyways, eh...I think this is OK as is.

cgwalters: (But if for example we ever wanted to use these tests on multi-arch, we'd need to handle substituting `$basearch` the same way libdnf is doing)

mike-nguyen:
Ok, I am merging this for now.  We will update if we get any new packages.

tests/improved-sanity-test/vars/common.yml
miabbott: > I think though this role falls into a gap where we're using rpm directly on top of ostree admin unlock 

Correct, that's the idea of running this particular role in the sanity test.

> (But if for example we ever wanted to use these tests on multi-arch, we'd need to handle substituting $basearch the same way libdnf is doing)

Shoot...that's going to be an issue in the not so distant future.  $DAYJOBBOSS actually just had a conversation with me about making our tests multi-arch aware.

PR #273, opened 2017-11-08
Add atomic CLI tests

mike-nguyen:
This commit adds tests for the the atomic command.  The tests revolve
mostly around pulling images and installing/running containers with
variations of the options in the atomic subcommand.  These tests include:

  - Pull by fully qualified image name
  - Pull by short image name
  - Pull by fully qualified image name with --name option
  - Pull with tags
  - Pull by digest
  - Pull to ostree/docker storage
  - Images update command

Most of the tests use the cockpit container.  This container was chosen
because it has most of the labels that the atomic command utilizes. For
simple tests, the busybox container was chosen because it is small.


mike-nguyen:
centos will likely fail because of https://github.com/projectatomic/docker/issues/285

roles/atomic_images_delete/tasks/main.yml
```diff
@@ -1,13 +1,14 @@
...
+  when: aid_image is undefined
+
+- name: Set options
+  set_fact:
+    del_options: "{{ aid_options | default() }}"
 
-- name: Remove specified image
-  command: atomic --debug --assumeyes images delete {{ image }}
+- name: Remove {{ aid_image }}
+  command: atomic --assumeyes images delete {{ aid_image }} {{ del_options }}
```
miabbott: nit: no newline at the end of the file (GitHub actually highlights this in the diff!)

tests/atomic/main.yml
```diff
@@ -0,0 +1,527 @@
...
+
+  pre_tasks:
+    - name: Set cockpit container name for RHELAH
+      set_fact:
+        cockpit_cname: "registry.access.redhat.com/rhel7/cockpit-ws"
+      when: ansible_distribution == "RedHat"
+
+    - name: Set cockpit container name for Fedora/CentOS
+      set_fact:
+        cockpit_cname: "registry.fedoraproject.org/f25/cockpit"
```
miabbott: Probably should bump this to F26; F27 should be released Nov 14 and that will EOL F25

tests/atomic/main.yml
```diff
@@ -0,0 +1,527 @@
...
+        image_name: "{{ cockpit_cname }}"
+      expect_missing: true
+
+    - role: atomic_install
+      ai_image: "{{ cockpit_cname }}"
+
+    - role: atomic_run
+      ar_image: "{{ cockpit_cname }}"
+
+    - role: atomic_images_list_verify
```
miabbott: Is it necessary to keep checking that the `cockpit` image is still there after each operation?

tests/atomic/main.yml
```diff
@@ -0,0 +1,527 @@
...
+  become: yes
+
+  tags:
+    - shortname
+
+  vars_files:
+    - vars.yml
+
+  pre_tasks:
+    - name: Set cockpit container name for RHELAH
```
miabbott: Maybe move these types of distro specific facts into vars files and use `with_first_found` to load them in?

tests/atomic/main.yml
```diff
@@ -0,0 +1,527 @@
...
+    - role: atomic_images_list_verify
+      expected_values:
+        repo: "{{ cockpit_fq_name }}"
+
+    - role: atomic_containers_list_verify
+      expected_values:
+        image_name: "{{ cockpit_short_name }}"
+      check_mode: false
+
+    - role: atomic_stop
```
miabbott: Might want to drop a comment here to explain the json queries and what you are looking for.

tests/atomic/main.yml
```diff
@@ -0,0 +1,527 @@
...
+        name: atomic_images_delete_all
+
+    - name: Pull {{ fq_name }} using ostree
+      include_role:
+        name: atomic_pull
+      vars:
+        apl_image: "{{ fq_name }}"
+        apl_options: "--storage=ostree"
+
+    # The TYPE field doesn't show up in the json output so
```
miabbott: Is this a bug in `atomic` itself?

miabbott:
Excellent work overall!  I like that you prefixed variable names in the roles to make them more unique per role (like we are supposed to do!)

I have a couple of minor comments, but I'm betting this can get merged very soon.  Would like to kick it around myself, though

tests/atomic/main.yml
mike-nguyen: I can drop this this check.

tests/atomic/main.yml
mike-nguyen: I forgot to file an upstream issue for that.  I'll add a comment in the code with the issue: https://github.com/projectatomic/atomic/issues/1129

tests/atomic/main.yml
mike-nguyen: I ended up using vars_files with the ansible_distribution name since there was no difference between the different versions of a distribution (ie fedora 26, 27, etc).  The only thing I had to account for the ansible_distribution for cahc which is centosdev.  If needed I could use include_vars with with_first_found.  I didn't feel that was needed for this simple case.

tests/atomic/main.yml
miabbott: Still looks like there a bunch of extra image checks in the test.  Could you audit the entire test and see if we really need each instance of `atomic_images_list_verify`?

miabbott:
`fedora/27/atomic` failure looks like a Fedora manifestation of https://bugzilla.redhat.com/show_bug.cgi?id=1498575

The build of `atomic` in F27 is from Aug 24!!!  Newer versions should have the fix for the problem observed.

tests/atomic/main.yml
```diff
@@ -0,0 +1,474 @@
...
+
+- name: Atomic Pull With Tags
+  hosts: all
+  become: yes
+
+  tags:
+    - pull_tags
+
+  pre_tasks:
+    - name: Set cockpit container name for RHELAH
```
miabbott: Is this comment correct, i.e. only for RHELAH?

miabbott:
Could we move the vars files under `tests/atomic/vars/` rather than have them all in the root directory?

mike-nguyen:
> Could we move the vars files under tests/atomic/vars/ rather than have them all in the root directory?

Should be possible.

mike-nguyen:
bot, retest this please

roles/atomic_containers_list/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+#
+# list containers and save to json
+#
+- name: List images with --json
+  command: atomic containers list --json
+  register: acl
+
+- name: Save JSON to variable
+  set_fact:
+    acl_json: "{{ acl.stdout | to_json | from_json }}"
```
miabbott: Is it necessary to use both filters?  Seems like you should just be able to use `from_json`?

roles/atomic_images_list/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  set_fact:
+    images_list_options: "{{ ail_options | default() }}"
+
+- name: List images with --json
+  command: atomic images list {{ images_list_options }} --json
+  register: ail
+
+- name: Save JSON to variable
+  set_fact:
+    ail_json: "{{ ail.stdout | to_json | from_json }}"
```
miabbott: Same question about both filters here

roles/docker_rmi/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+#
+#  Removes an image
+#
+
+- name: Fail if drmi_image
+  fail:
+    msg: "drmi_image is not defined"
+  when: drmi_image is undefined
+
+- name: Set optioins
```
miabbott: `s/optioins/options/`

roles/docker_tag/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: Fail if dt_image or dt_tag is not defined
+  fail:
+    msg: "dt_image or dt_tag is not defined"
+  when: dt_image is undefined or
+        dt_tag is undefined
+
+- debug:
+    msg: "dt_img: {{ dt_image }}"
+
+- debug:
```
miabbott: Did you want to leave these `debug:` statements here?

mike-nguyen: Nope, removed!

roles/atomic_containers_list/tasks/main.yml
mike-nguyen: I don't need both.  I'm not quite sure why I did that.

mike-nguyen:
New fixup with the comments addressed.  I also snuck in a fix to handle the new TOML /etc/containers/registries.conf

roles/docker_private_registry/tasks/main.yml
```diff
@@ -14,13 +14,25 @@
...
+    when: reg.stat.exists
+
+  - name: add toml registeries
+    replace:
+      regexp: "\\[registries.insecure\\]\nregistries = \\[\\]"
+      replace: "[registries.insecure]\nregistries = ['{{ ansible_docker0.ipv4.address }}:5000']"
+      dest: /etc/containers/registries.conf
+    when: reg.stat.exists and toml.rc == 0
+
   - name: add registeries
```
miabbott: s/registeries/registries/

mike-nguyen: https://i.pinimg.com/736x/3a/11/68/3a116884945f870924f1ffd3f36fc015--los-simpsons-homer-simpson.jpg

roles/docker_private_registry/tasks/main.yml
```diff
@@ -14,13 +14,26 @@
       path: /etc/containers/registries.conf
     register: reg
 
-  - name: add registeries
+  - name: Check for TOML formmated registries.conf
+    command: grep registries.search /etc/containers/registries.conf
+    register: toml
+    when: reg.stat.exists
+    ignore_errors: true
+
```
mike-nguyen: I originally forgot to ignore errors on the grep so it failed when it wasn't a TOML file in the case of fedora/26/atomic.  Fixed it here.

mike-nguyen:
The CentOS failures are due to https://github.com/projectatomic/docker/issues/285.  CentOS won't use the fixed version of docker until https://github.com/projectatomic/docker/issues/286 is fixed.  The current version of docker on CentOS is using a version of docker that doesn't have a /etc/containers/registries.conf which causes the test to fail because the test uses an insecure registry.  Once the docker fix is in CentOS the tests should pass.

PR #275, opened 2017-11-09
roles: adjust retries on 'redhat_subscription' and 'redhat_unsubscribe'

miabbott:
Our internal jobs hit some network timeouts while doing an
'unsubscribe' with `subscription-manager`.  We already had some
retries for `redhat_subscription`, but were lacking them on the
`redhat_unsubscribe`.  This change implements the retries for
'unsubscribe' and bumps up the retries on 'subscription'


mike-nguyen:
LGTM

PR #276, opened 2017-11-15
papr: changes related to f27 release

miabbott:
F27 has been released, so let's enable our CI on that image and switch
the containers to F27, too.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -3,7 +3,7 @@ cluster:
     - name: testnode
       distro: fedora/26/atomic
   container:
-    image: registry.fedoraproject.org/fedora:25
+    image: registry.fedoraproject.org/fedora:27
 
 context: fedora/26/atomic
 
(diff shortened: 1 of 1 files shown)


miabbott:
In the spirit of spreading our review coverage around, @jlebon do you want to have peek?

.papr.yml
```diff
@@ -3,7 +3,7 @@ cluster:
     - name: testnode
       distro: fedora/26/atomic
   container:
-    image: registry.fedoraproject.org/fedora:25
+    image: registry.fedoraproject.org/fedora:27
```
jlebon: I'd just keep 26 here to make sure running from the f26 userspace is still OK. We're already testing 27 down there right?

miabbott: In the same vein, does it make sense to switch the `centos/7/atomic` contexts to use a CentOS base image?

jlebon: I'd say so yeah. More so than f26 vs f27. :)

jlebon:
LGTM! 👍

PR #277, opened 2017-11-15
README: updates for f27, etc

miabbott:
We will soon have test results for the
`fedora/27/x86_64/updates/atomic-host` ref, so update our dashboard
with the necessary info.

Additionally, I cleaned up the 'supported tests' section.


jlebon:
LGTM.

PR #278, opened 2017-11-17
Add test for livefs

mike-nguyen:
This commit adds a simple livefs test using package layering as the
change in the filesystem.

The rpm_ostree_install_verify and rpm_ostree_uninstall_verify roles
needed an option to skip the rpm-ostree status since the output is
changed with the livefs implementation. The roles were also cleaned
up and now uses the new variables from the rpm_ostree_status role.

### DIFF
--- a/roles/rpm_ostree_livefs/tasks/main.yml
+++ b/roles/rpm_ostree_livefs/tasks/main.yml
@@ -1,0 +1,19 @@
+---
+# vim: set ft=ansible:
+#
+- name: Enable livefs
+  command: rpm-ostree ex livefs
+  register: ros_lfs
+
+# Regex to find all the sha256 checksums in the output
+- name: Get commits
+  set_fact:
+    commits: "{{ ros_lfs.stdout|regex_findall('[A-Fa-f0-9]{64}') }}"
+
+# Note: There was an instance where there were three checksums in the message
+# so just grab the last two.
+- name: Get base and livefs commits
+  set_fact:
+    rol_base_commit: "{{ commits[-2] }}"
+    rol_livefs_commit: "{{ commits[-1] }}"
+
--- a/roles/rpm_ostree_install_verify/tasks/main.yml
+++ b/roles/rpm_ostree_install_verify/tasks/main.yml
@@ -9,6 +9,7 @@
 # Parameter:
 #   roiv_package_name - name of a single package
 #   roiv_binary_name - name of binary to check for (optional)
+#   roiv_status_check - boolean - optionally check the rpm-ostree status for package
 #
 # Requirements:
 #  This role must be run after the reboot following rpm-ostree install of a package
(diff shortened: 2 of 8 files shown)


roles/rpm_ostree_livefs/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+# vim: set ft=ansible:
+#
+- name: Enable livefs
+  command: rpm-ostree ex livefs
+  register: ros_lfs
+
+# Regex to find all the sha256 checksums in the output
+- name: Get commits
+  set_fact:
+    commits: "{{ ros_lfs.stdout|regex_findall('[A-Fa-f0-9]{64}') }}"
```
cgwalters: This data is also in `rpm-ostree status --json` (key `live-replaced` etc.); scraping it is going to be fragile.  There's code in some of the other tasks that parse that right?

mike-nguyen: I'm using the checksum from the output to verify the live-replaced in the rpm-ostree status output.  Is there a way to generate the checksum directly so I can use it to verify the livefs output and rpm-ostree status output?

mike-nguyen:
bot, retest this please

mike-nguyen:
The CI tests are timing out for some reason.  I tried running the test locally and ran into an issue.  For some reason a variable I am setting is not getting set and causing a failure when running the whole test suite.  When running the single play, it works.  I will investigate this and update the PR with any findings.

roles/rpm_ostree_livefs/tasks/main.yml
miabbott: I think it would be reasonable to use `rpm-ostree status --json` and then just pull the commit data we want from there:

```
- name: Enable livefs
  command: rpm-ostree ex livefs
  register: ros_lfs

- name: Get status
  command: rpm-ostree status --json
  register: ros

- name: Convert to JSON
  set_fact:
    ros_json: "{{ ros.stdout | from_json }}"

- name: Query for commits
  set_fact:
    rol_base_commit: "{{ ros_json['deployments'][0]['base-checksum'] }}"
    rol_livefs_commit: "{{ ros_json['deployments'][0]['checksum'] }}"
```

This might be a little fragile because we are assuming that deployment 0 is the pending deployment that was made live, but we should have enough confidence in the state of the deployments when using this role.

miabbott:
This is looking good.  Will wait on resolution to the comments made before going forward with merging.

roles/rpm_ostree_livefs/tasks/main.yml
miabbott: After talking to @mike-nguyen on IRC, I'm going to retract my suggestion here.  

The intent is to compare the commit IDs from the output of `rpm-ostree ex livefs` to the output from `rpm-ostree status`, so my suggestion doesn't quite fit.

miabbott:
I believe all comments have been addressed, so going to merge this.

Issue #262, opened 2017-10-03
change PAPR/test_director to use virtualenv


miabbott:
Our current PAPR definition is installing Ansible and other packages via `dnf`.  This has been landing a version of Ansible (2.3 or newer) which we haven't committed supporting yet.

We should change our PAPR definition to setup a virtualenv and land a version of Ansible that we support with our tests.

miabbott:
Closed via #279

PR #279, opened 2017-11-17
Use virtualenv for PAPR

mike-nguyen:
Resolves issue #262.  The latest version of Ansible was being installed
by PAPR but A-H-T currently supports 2.2.x.  Use virtualenv to install
the supported version of Ansible for A-H-T.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -8,9 +8,12 @@ cluster:
 context: fedora/26/atomic
 
 packages:
-  - ansible
+  - python-virtualenv
+  - python-pip
   - git
-  - python2-jmespath
+  - gcc
+  - redhat-rpm-config
+  - libselinux-python
 
 tests:
   - ./.test_director
@@ -23,7 +26,7 @@ cluster:
     - name: testnode
       distro: fedora/27/atomic
   container:
-    image: registry.fedoraproject.org/fedora:27
+    image: registry.fedoraproject.org/fedora:26
 
 context: fedora/27/atomic
 
@@ -35,7 +38,7 @@ cluster:
     - name: testnode
       distro: centos/7/atomic
   container:
-      image: registry.centos.org/centos/centos:7
+      image: registry.fedoraproject.org/fedora:26
 
 context: centos/7/atomic
 
@@ -47,6 +50,6 @@ cluster:
     - name: testnode
       distro: centos/7/atomic/continuous
   container:
-      image: registry.centos.org/centos/centos:7
+      image: registry.fedoraproject.org/fedora:26
 
 context: centos/7/atomic/continuous
--- a/.test_director
+++ b/.test_director
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -xeuo pipefail
 
+VIRTUAL_ENV_DISABLE_PROMPT=1
+
+if [ ! -d venv ]; then
+    virtualenv --system-site-packages venv
+fi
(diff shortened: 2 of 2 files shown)


mike-nguyen:
Slapping a WIP label on it as I test it out on this environment.  I will remove it once I have something working.

mike-nguyen:
bot, retest this please

mike-nguyen:
VIRTUAL_ENV_DISABLE_PROMPT=1 doesn't seem to work when running from CentOS containers.  @jlebon @miabbott have any idea why?

Is anyone opposed to running all the CI tests from a Fedora container?

miabbott:
@mike-nguyen Maybe try using the `env` parameter - https://github.com/projectatomic/papr/blob/master/docs/sample.papr.yml#L136

jlebon:
Yeah, I've hit that. I think the activate script changed recently and it's referring to a bunch of new other variables. So we'd have to define those ahead of time too. But this approach is bound to relapse the next time it changes again. Looks like there's an issue upstream to track this already: https://github.com/pypa/virtualenv/issues/1029.

I would just bite the bullet and turn off `set -u` for the activation phase at least. Or maybe just make it invoke the real shell script in which you turn `set -u` back on?

miabbott:
This comment has a pretty clean workaround, IMO

https://github.com/pypa/virtualenv/issues/1029#issuecomment-341489487

mike-nguyen:
> This comment has a pretty clean workaround, IMO
> pypa/virtualenv#1029 (comment)

@miabbott Are you talking about `VIRTUAL_ENV_DISABLE_PROMPT=true source env/bin/activate`?
I already have that in there but its just not working on CentOS.  I think I might have to go with @jlebon's suggestion and turnoff set -u and turn it back on after activation.

miabbott:
@mike-nguyen I think I parsed that comment I linked too quickly.  I imagined doing something like this in `.papr.yml`:

```
- tests:
    - VIRTUAL_ENV_DISABLE_PROMPT=1 ./.test_director
```

But that probably wouldn't fly.

miabbott:
The two CentOS failures look like network flakes trying to get in touch with PyPi...

.papr.yml
```diff
@@ -8,9 +8,11 @@ cluster:
 context: fedora/26/atomic
 
 packages:
-  - ansible
+  - python-virtualenv
```
miabbott: Let's add `libselinux-python` to this list, which is used when Ansible writes out files to the host.  We probably don't need this for PAPR, but it wouldn't hurt.

.test_director
```diff
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -xeuo pipefail
 
+VIRTUAL_ENV_DISABLE_PROMPT=1
+
+if [ ! -d venv ]; then
+    virtualenv venv
```
miabbott: With `libselinux-python` installed, we should use `virtualenv --system-site-packages venv` to ensure we get the SELinux bindings available.

.papr.yml
mike-nguyen: I added `libselinux-pytho`n and also the ` --system-site-packages` option.  Looks to be working from the checks passing!

mike-nguyen:
Removed WIP label.  This is ready to be reviewed.

miabbott:
Looks good!

PR #280, opened 2017-11-17
admin-unlock: use different RPM on Fedora

miabbott:
In #271, the RPM used during the sanity test of `ostree admin unlock`
was changed for the Fedora case.  We make the same changes here in the
more comprehesive test of the `ostree admin unlock` functionality.

As part of this, the base EPEL URL is updated (who knew it chagned?!)
and the `sed` regex was updated to be a little cleaner.

### DIFF
--- a/roles/epel_rpm_url/tasks/main.yml
+++ b/roles/epel_rpm_url/tasks/main.yml
@@ -21,14 +21,14 @@
 
 - name: Set EPEL base url
   set_fact:
-    epel_base: "https://download.fedoraproject.org/pub/epel/7/x86_64"
+    epel_base: "https://download.fedoraproject.org/pub/epel/7/x86_64/Packages/"
 
 - name: Get subdir url
   set_fact:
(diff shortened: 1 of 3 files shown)


miabbott:
This is not super clean to look at, but it works.

Longer term, I think the better solution is to create a dummy RPM that drops a file on the host and use that for our tests.

jlebon:
Here's a simple RPM as well as its spec file and Makefile: https://jlebon.fedorapeople.org/aht-dummy/. You can do `make rpm` to recreate the RPM if you need to modify the spec. Maybe let's just import these files into the repo and install the RPM directly?

miabbott:
I'm going to mark this as WIP since @jlebon has graciously provided us an RPM to use.  Might close this PR in favor of a larger one that also changes up the `improved-sanity-test`

mike-nguyen:
@jlebon :+1: :+1:  :+1:

miabbott:
Hoo boy!  The dummy RPM route is way easier than this `when:` nightmare.

Closing this in favor of #281 and #282

PR #281, opened 2017-11-20
add files to support a 'dummy' RPM

miabbott:
We have a need for an RPM that can be installed across all platforms
(Fedora, RHEL, CentOS) and @jlebon was kind enough to whip up the
necessary pieces to create said RPM.  This brings those pieces under
source control.


mike-nguyen:
Can we put these files in a sub-directory instead of on the root of atomic-host-tests?

miabbott:
Moved in the fixup ⬆️

mike-nguyen:
LGTM

PR #282, opened 2017-11-20
rpm_install: use dummy RPM to support cross-platform installs

miabbott:
Previously we had been able to use the `epel-release` RPM as a way to
test RPM installs when a host had used `ostree admin unlock`.  But the
`epel-release` RPM was changed to conflict with the `fedora-release`
RPM, meaning we were not able to use a single RPM across all
platforms.

Instead of trying to setup conditions for when to use `epel-release`
and when not to, we are now able to use a 'dummy' RPM that will
install across all platforms.  (Thanks @jlebon!)

This change implements the use of the 'dummy' RPM in the tests that
use the `rpm_install` role.

Additionally, I snuck in some minor changes to the `epel_rpm_url`
role:  updating the base EPEL URL and improving the `sed` regex for
readability.

Depends on #281


tests/improved-sanity-test/vars/common.yml
```diff
@@ -7,10 +7,8 @@ g_file2: 'commit2'
 g_dir2: 'qe2'
 g_hostname: 'myhostname'
 g_pkg: "wget"
-g_rpm_url: 'https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm'
-g_rpm_url_fedora: 'https://kojipkgs.fedoraproject.org/packages/ltrace/0.7.91/22.fc26/x86_64/ltrace-0.7.91-22.fc26.x86_64.rpm'
-g_rpm_name: 'epel-release'
-g_rpm_name_fedora: 'ltrace'
+g_rpm_url: 'https://raw.githubusercontent.com/projectatomic/atomic-host-tests/master/aht-dummy-1.0-1.noarch.rpm'
```
jlebon: That's cool, though we could copy over and install the RPM from our checkout, right? One less thing that relies on external infra. :)

miabbott: We could...but the `rpm_install` role expects a URL to install from.

I could rework the role to handle either a discrete file or a URL...let me investigate if it would be worthwhile.

miabbott:
Bleh, had to force push my last commit for some reason  😞

miabbott:
Oops, left some commented out pieces in there.  Will get a fixup soon.

roles/epel_rpm_url/tasks/main.yml
```diff
@@ -21,14 +21,14 @@
...
-    epel_base: "https://download.fedoraproject.org/pub/epel/7/x86_64"
+    epel_base: "https://download.fedoraproject.org/pub/epel/7/x86_64/Packages"
 
 - name: Get subdir url
   set_fact:
     epel_subdir: "{{ epel_base }}/{{ rpm_name[0] }}"
 
 - name: Scrape HTML for package name
-  shell: curl -LsSk {{ epel_subdir }}/ | sed -n 's/.*"\({{ rpm_name }}.*.rpm\)\".*/\1/p'
+  shell: curl -LsSk {{ epel_subdir }}/ | sed -n 's|.*\"\({{ rpm_name | quote }}.*.rpm\)\".*|\1|p'
```
miabbott: Hmph...missing another `quote` filter here on the `epel_subdir` variable

miabbott:
I ran the `centos/7/atomic` and `fedora/27/atomic` locally and they passed.  Not sure why the connection is getting dropped in the PAPR tests.

mike-nguyen:
bot, retest this please

mike-nguyen:
LGTM, just re-testing for a clean CI check since you tested locally and it worked.  Hopefully it will work here when the checks re-run.

mike-nguyen:
bot, retest this please

roles/epel_rpm_url/tasks/main.yml
```diff
@@ -21,14 +21,14 @@
...
 - name: Get subdir url
   set_fact:
     epel_subdir: "{{ epel_base }}/{{ rpm_name[0] }}"
 
 - name: Scrape HTML for package name
-  shell: curl -LsSk {{ epel_subdir }}/ | sed -n 's/.*"\({{ rpm_name }}.*.rpm\)\".*/\1/p'
+  shell: curl -LsSk {{ epel_subdir | quote }}/ | sed -n 's|.*\"\({{ rpm_name | quote }}.*.rpm\)\".*|\1|p'
   register: epel_output
   retries: 5
   delay: 60
```
mike-nguyen: I noticed in the CI failures, the curl gets a 404 and sed no-ops but still results in a 0 return code and the test fails further down the line.  Maybe we should split this into two tasks--one to curl the epel subdir (retry until we don't get a 404), then one to grab the rpm name from the output?

miabbott: Seems reasonable, I'll look into it.

Although, the constant `curl` failures when talking to the Fedora mirrors is both troublesome and annoying.

mike-nguyen: I agree its pretty annoying.  For some reason when I curl repeatedly on my local system, it worked every time.

miabbott:
Fixup with separation of network access and regex matching ⬆️

miabbott:
Hmph...the regex worked locally for me, but is failing the tests....back to the drawing board...

miabbott:
I think some mirrors are better than others and the output scraped from them would usually work.  But when it doesn't, the error is opaque and I'm left scratching my head.

I'm tired of fighting with the mirrors.  The latest fixup goes straight to the source; no mirrors involved.  ⬆️

miabbott:
fuck yeah

miabbott:
@mike-nguyen notes that we stopped using `dl.fedoraproject.org` in #177 because it was down at one point and caused all of our tests to fail.  But using the mirror network here was giving me multiple failures in the CI checks (maybe due to infra problems in the backend), so in reality we can't win either way.

mike-nguyen:
Given the flakiness of the mirrors, I think it would be much easier to triage failures if everything is failing due to one site being down than random failures due to a particular mirror being down.  I'm good with the state of the current PR unless someone disagrees.

mike-nguyen:
I'm going to merge this and we can just revert if we have issues with the direct link to the rpm.

PR #283, opened 2017-12-04
stop using Fedora 25 images in roles/tests

miabbott:
Fedora 25 will be EOL shortly, so let's bump up to Fedora 26.  (I
would use Fedora 27, but not all of the old images got rebuilt, AFAIK)

### DIFF
--- a/roles/atomic_installation_verify/tasks/main.yml
+++ b/roles/atomic_installation_verify/tasks/main.yml
@@ -19,7 +19,7 @@
 
 - name: Set cockpit container name for Fedora/CentOS
   set_fact:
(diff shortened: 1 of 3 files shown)


mike-nguyen:
LGTM

miabbott:
The `fedora/26/atomic` is a failure due to it using an older version of `atomic` where it tried to install an RPM by default when doing `atomic install --system`.  I'll submit another PR to tackle this condition.

mike-nguyen:
bot, retest this please

miabbott:
nailed it

PR #285, opened 2017-12-04
Increase likelihood that roles run more than once with allow_duplicates

mike-nguyen:
There were a few instances while running atomic host tests where
where a role was not being run a second time (Ansible's default
behavior).  This commit addes meta/main.yml for each role with
allow_duplicates set to true to increase the likelihood that the
role will run when it is subsequently called.

check_mode: no is also another method and can be used if this doesn't
work.


miabbott:
Awesome!

PR #286, opened 2017-12-08
Gather version information for common packages

mike-nguyen:
The goal of atomic host tests is to run across the latest released
and development streams of Atomic Host (Fedora, RHEL, CentOS). Some
of these streams are slower to get newer packages and causes breakage
in the test due to missing/broken functionality.  We have dealing
with these by getting the version of the package as needed and
conditionally performing actions based on the version.

This PR aims to simplify this process by centralizing the gathering
of about the atomic host in one role and making the information
available through a dictionary.  Currently, there is only version
information about common packages, but anything can be added to the
dictionary.  It is pretty much like the setup module in Ansible but
specifically for Atomic Host.

### DIFF
--- a/roles/atomic_host_set_facts/meta/main.yml
+++ b/roles/atomic_host_set_facts/meta/main.yml
@@ -1,0 +1,2 @@
+---
+allow_duplicates: true
--- a/roles/atomic_host_set_facts/tasks/main.yml
+++ b/roles/atomic_host_set_facts/tasks/main.yml
@@ -1,0 +1,39 @@
+---
+# vim: set ft=ansible:
+#
+# This role creates a dictionary named atomic_host that contains the versions
+# of common packages contained in atomic host.
+#
+- name: Get atomic version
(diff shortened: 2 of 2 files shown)


mike-nguyen:
Since all the version gathering is the same command, I was trying to think of a way to get it into a loop.  In Ansible, it seems like I need a two step process to grab the data, then to set it as a fact.  I was thinking I could use a loop with an include.  If you have any ideas I would be open to it.  I will give it a try and update this PR.

mike-nguyen:
WIP label as I work out whether this is the best approach.

miabbott:
I wonder if this could be done via dictionary?  Having multiple `set_fact` statements seems cumbersome.

This SO post might be relevant

https://stackoverflow.com/questions/35605603/using-ansible-set-fact-to-create-a-dictionary-from-register-results

Also, `openshift-ansible` does a lot of fact gathering...might be worth while to look at their project and see how they do it.

mike-nguyen:
Abandoning this PR and going a different direction.

PR #284, opened 2017-12-04
tests: handle 'atomic install --system' on old versions

miabbott:
Well, we're still using `fedora/26/atomic` in our CI checks and the
version of `atomic` there is friggin old.  How old?  So old that it
will try to land an RPM on the host when you do an 'atomic install
--system' by default.

This tries to work around this problem by doing a version check and
altering how the `atomic install` command is used depending on the
version detected.

Issue #239, opened 2017-08-25
perform audit of roles to print expected values when doing comparisons


miabbott:
As pointed out in #238, some of our roles are doing value comparisons, but the error messages do not include the expected values.

We should do an audit of the roles to make sure we are including the expected values in any error messages that are printed.

PR #288, opened 2017-12-11
Improve error messages in roles

mike-nguyen:
Sometimes its not easy to figure out why a test failed based on
the error message.  Digging into the code to find the failure
comparison that caused the failure is required.

Lets make the error messages more detailed when comparisons are
made so triaging failures is easier.

Note: This commit is only improving the error messages in the roles.

### DIFF
--- a/roles/atomic_host_status_verify/tasks/main.yml
+++ b/roles/atomic_host_status_verify/tasks/main.yml
@@ -19,5 +19,8 @@
 
 - name: Fail if output is not identical
   fail:
-    msg="The output of atomic host status and rpm-ostree status are not the same"
+    msg: |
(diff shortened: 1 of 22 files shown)


mike-nguyen:
fedora/26/atomic failure is unrelated to this PR

miabbott:
This looks like it will mostly address #239, right?

mike-nguyen:
> This looks like it will mostly address #239, right?
Yes, I will do the actual tests next.

miabbott:
I tried to get the `fedora/26/atomic` context to pass, but it looks like we are hitting the same condition that was addressed in #284 

I'll submit another PR for that issue, but this one is good to merge!  Thanks for cleaning this up.

PR #289, opened 2017-12-11
add 'atomic_install_version_check' role

miabbott:
I was about to add a 3rd copy of the same code to detect which version
of `atomic` we are using when we run the tests.  But I came to my
senses and added a new role that handles all that.

The `atomic_install_version_check` role does a version check and sets
up a variable to use `atomic install`.  This is a follow-on (and
improvement, IMO) to #284.

### DIFF
--- a/tests/system-containers/main.yml
+++ b/tests/system-containers/main.yml
@@ -391,24 +391,9 @@
     - vars.yml
 
   pre_tasks:
-    - name: Determine version of atomic
-      command: rpm -q --queryformat '%{VERSION}' atomic
-      register: rpmq
-
-    - name: Setup atomic_version fact
-      set_fact:
-        atomic_version:  "{{ rpmq.stdout }}"
-
-    # The behavior changed with this commit:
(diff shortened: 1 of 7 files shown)


tests/system-containers/main.yml
```diff
@@ -465,7 +452,7 @@
 
     - name: Install flannel
       command: >
-        {{ atomic_install_cmd }}
+        {{ aivc_atomic_install_cmd }}
```
mike-nguyen: Maybe add the same comment above this statement about where aivc_atomic_install_cmd comes from?  I'm not sure if its really necessary since its mentioned above but might help someone looking at the code.

miabbott:
> Side note: I wonder if we can figure out a way to get #286 worked into this somehow.

I think we can roll-up this change into #286 eventually

miabbott:
More comments never hurt - fixup pushed ⬆️

mike-nguyen:
Great work as usual!

PR #290, opened 2017-12-12
Improve error messages in tests

mike-nguyen:
This is the follow on to #288 to address #239.  #288 addressed the
error messages in the roles and this commit addresses the error
messages in the tests themselves.

### DIFF
--- a/tests/admin-unlock/main.yml
+++ b/tests/admin-unlock/main.yml
@@ -235,7 +235,9 @@
 
     - name: Fail if current deployment has unlocked not set to none
       fail:
-       msg: "Booted deployment does not have 'unlocked' not set to none"
+       msg: |
+         Expected: booted deployment has unlocked set to none
+         Actual: booted deployment unlocked is set to {{ ros_booted['unlocked'] }}
       when: "'none' not in ros_booted['unlocked']"
 
 - name: Ostree Admin Unlock - Unlock Twice - Error Message
@@ -257,7 +259,9 @@
 
     - name: Verify rpm_ostree status is unlocked and set to hotfix
       fail:
-        msg: "Booted deployment does not have 'unlocked' set to hotfix"
+        msg: |
+          Expected: booted deployment has unlocked set to hotfix
+          Actual: booted deployment unlocked is set to {{ ros_booted['unlocked'] }}
       when: "'development' not in ros_booted['unlocked']"
 
     - name: Run unlock twice (system is already unlocked)
@@ -267,7 +271,10 @@
 
     - name: Fail when error message is incorrect
       fail:
-        msg: "Error message is incorrect"
+        msg: |
+          Expected: Error message should indicated the deployment is already in
+                    the unlocked state: development
+          Actual: double_unlock.stderr
       when: "'Deployment is already in unlocked state: development' not in double_unlock.stderr"
 
     - name: Run unlock hotfix (system is already locked)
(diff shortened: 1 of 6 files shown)


mike-nguyen:
bot, retest this please

mike-nguyen:
I just realized that this is going to run every test we have and most likely it will timeout/fail.

tests/admin-unlock/main.yml
```diff
@@ -267,7 +271,10 @@
 
     - name: Fail when error message is incorrect
       fail:
-        msg: "Error message is incorrect"
+        msg: |
+          Expected: Error message should indicated the deployment is already in
+                    the unlocked state: development
+          Actual: double_unlock.stderr
```
miabbott: Do we need double curly braces here?  It probably works without them, but adding them would be consistent elsewhere.

tests/pkg-layering/main.yml
```diff
@@ -310,7 +318,10 @@
 
     - name: Fail if error not expected
       fail:
-        msg: "Command error not expected"
+        msg: |
+          Expected: {{ g_invalid_pkg }} is not currently requested in rpm-ostree
+                    uninstall output
+          Actual: {{ uninstall.stderr }}
       when: "'\\'{{ g_invalid_pkg }}\\' is not currently requested' not in uninstall.stderr"
```
miabbott: As an aside unrelated to the changes here, this conditional looks ugly as sin.  I wonder if we can improve it....

tests/admin-unlock/main.yml
mike-nguyen: I missed that one!  I think it would actually print out the literal double_unlock.stderr

tests/system-containers/main.yml
```diff
@@ -353,12 +369,16 @@
 
     - name: Fail when rootfs dir is missing from main container
       fail:
-        msg: "/var/lib/containers/atomic/{{ g_hw_name }}/rootfs does not exist"
+        msg: |
+          Expected: /var/lib/containers/atomic/{{ g_hw_name }}/rootfs exists is True
+          Actual: /var/lib/containers/atomic/{{{ g_hw_name }}/rootfs exists is {{ hw.stat.exists }}
```
miabbott: extra opening curly brace

mike-nguyen: I'm surprised Ansible didn't choke on this when I tested the test

tests/rpm-ostree/main.yml
```diff
@@ -751,17 +751,23 @@
...
+          Expected: -I in initramfs args
+          Actual:  {{ ros_booted['initramfs-args'][0] }}
       when: "'-I' not in ros_booted['initramfs-args'][0]"
 
     - name: Fail if second initramfs argument is not set
       fail:
-        msg: "Initramfs args are incorrect.  Expected /etc/rpmostree-file.  Actual {{ ros_booted['initramfs-args'][1] }}"
+        msg: |
+          Expected: /etc/rpmostree-file in rpm-ostree initramfs-args
+          Actual: {{ ros_booted['initramfs-args'][1] }}"
```
miabbott: extra double quote at the end of the line

miabbott:
Looks good overall.  Just a couple of nits and we can probably merge this in.

miabbott:
None of the CI failures are due to the change itself, so this is good to merge.  Thanks for continuing the cleanup!

PR #291, opened 2017-12-15
Add RPM version retrieval

mike-nguyen:
Sometimes functionality may differ in an RPM released in one atomic
host versus another.  Conditionals must be setup to handle the
differences.  Until now, this was done in the individual test/role.
This commit centralizes that functionality into one role so the
test can reference the same dictionary to get RPM version information.

rpm_version is a base role that will get the RPM versions from a list
of RPMs.  atomic_host_versions gathers version information for a default
set of RPMS (or user specified set of RPMs) and can also optionally fail
if minimum versions of RPMs are not met.

Removed runc_version_check role because it is not used and obsoleted
by the addition of these roles.


roles/rpm_version/tasks/version.yml
```diff
@@ -0,0 +1,22 @@
...
+  when: rv_rpm is undefined
+
+- name: Determine version of {{ rv_rpm }}
+  command: rpm -q --queryformat '%{VERSION}' {{ rv_rpm }}
+  register: rpmq
+
+- name: Set facts
+  set_fact:
+    rv_version: "{{ rpmq.stdout }}"
+    atomic_host: "{{ atomic_host|default({}) | combine( { rv_rpm: rpmq.stdout } ) }}"
```
miabbott: This this variable is going to be globally available, it should probably be prefaced with `g_`

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -2,119 +2,117 @@
...
-  - name: pull kubernetes api-server, controller-mgr, scheduler
-    command: docker pull {{ item }}
-    register: pull
-    failed_when: item not in pull.stdout
-    with_items:
-      - "{{ kube_apiserver }}"
-      - "{{ kube_controller_manager }}"
-      - "{{ kube_scheduler }}"
+- name: make db directory
+  file: path=/etc/kubernetes/manifests state=directory mode=0755
```
miabbott: Since this role is getting modified, do you want to change this task to use the format found elsewhere?

```
- name: make db directory
  file:
    path: /etc/kubernetes/manifests
    state: directory
    mode: 0755
```

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -2,119 +2,117 @@
...
 
-  - name: make db directory
-    file: path=/etc/kubernetes/manifests state=directory mode=0755
+- name: template out apiserver, controller-manager, and scheduler pods
+  template:
+    src=roles/kubernetes_setup/files/{{ item }}.j2
+    dest=/etc/kubernetes/manifests/{{ item }}.json
+    owner=root
+    group=root
+    mode=0644
```
miabbott: Same recommendation as made on line #37

roles/atomic_host_versions/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  vars:
+    rv_rpms: "{{ ahv_missing_rpms }}"
+  when: ahv_missing_rpms|length > 0
+
+- name: Fail if version requirement is not met
+  fail:
+    msg: |
+      Expected: {{ item.key }} is running version {{ ahv_requirements[item.key] }}
+      Actual: {{ item.key }} is running {{ atomic_host[item.key] }}
+  when: atomic_host[item.key] | version_compare(item.value, '<') and atomic_host[item.key] is defined
```
miabbott: My preference is to use a list for logical ANDs in `when:` statements, but this works fine.

For example:

```
when:
  - atomic_host[item.key] | version_compare(item.value, '<')
  - atomic_host[item.key] is defined
```

roles/kubernetes_setup/tasks/main.yml
mike-nguyen: Deleted two spaces from the beginning of each line now I have to do all the cleanup work. haha

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -2,119 +2,117 @@
...
-      - kubelet.service
+- name: start docker, kube-proxy.service, kubelet.service
+  service:
+    name={{ item }}
+    state=started
+    enabled=yes
+  with_items:
+    - docker
+    - kube-proxy.service
+    - kubelet.service
```
miabbott: Same recommendation here to switch from `param=value` to `param: value`

roles/rpm_version/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# vim: set ft=ansible:
+#
+# Parameters:
+#   rv_rpms (required) - list of rpms installed on atomic host
+#
+# This role takes a list of rpms, retrieves the version, and stores it in a
+# dict named atomic_host
+#
+- include: version.yml rv_rpm={{ item }}
```
miabbott: Another place to switch to `param: value`

miabbott:
Great idea to consolidate all this.  Just some minor nits, but this is overall good.

miabbott:
Woo CI green means merge merge merge

PR #292, opened 2017-12-15
docker-build-httpd:  set MTU equal to default interface

miabbott:
When running an OpenStack environment, the host under test might end
up with an MTU that doesn't jive with the Docker daemon (usually when
the MTU is lower than 1500).  This can lead to network issues when
doing `docker build`.  So we'll just set the MTU for the Docker
daemon to the same.

### DIFF
--- a/tests/docker-build-httpd/main.yml
+++ b/tests/docker-build-httpd/main.yml
@@ -39,6 +39,17 @@
       when: (rhel_base_images is undefined or rhel_image_names is undefined) and
             ansible_distribution == "RedHat"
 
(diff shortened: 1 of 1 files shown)


jlebon:
I'd add a link to https://bugzilla.redhat.com/show_bug.cgi?id=1475460 somewhere in there. :)

miabbott:
Huh...need to install `rsync` in the container running the test.  See #294

mike-nguyen:
bot, retest this please

mike-nguyen:
LGTM

PR #296, opened 2017-12-18
Small improvements to i-s-t

ashcrow:


mike-nguyen:
LGTM.  I like the when statements in the beginning!  It reads more like an if statement and easier to follow the logic.

miabbott:
Thanks for the extra comments; definitely helps when reading the playbook!

PR #301, opened 2017-12-20
WIP:  use 'jq' to filter out noisy keys from JSON output

miabbott:
In the newer versions of `rpm-ostree`, the RPM DB package list is
included in the commit metadata.  This causes the output of
`rpm-ostree status --json` to be very noisy when run via Ansible (i.e.
100s of package key/values are now in the output).

This attempts to work around the noise by running the output through
`jq` and deleting the offending keys.

Part of this included creating a new role to land `jq` on the host, if it was not already there.

### DIFF
--- a/roles/install_jq/defaults/main.yml
+++ b/roles/install_jq/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+jq_path: /usr/local/bin/jq
+jq_url: https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64
--- a/roles/install_jq/meta/main.yml
+++ b/roles/install_jq/meta/main.yml
@@ -1,0 +1,2 @@
+---
+allow_duplicates: true
--- a/roles/install_jq/tasks/main.yml
+++ b/roles/install_jq/tasks/main.yml
@@ -1,0 +1,16 @@
+---
+# vim: set ft=ansible:
+#
+- name: Test if 'jq' is installed
+  command: command -v jq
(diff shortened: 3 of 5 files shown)


miabbott:
I need to make the `rpm_ostree_status` role use the version of `rpm-ostree` to determine when to filter the output through `jq`, but it is 4pm on the day before holiday break and I want to go have a beer.

cgwalters:
Seems OK, though down the line we should start thinking about changing a-h-t to use a "tools" container for some things; I think at least openshift-ansible is doing this?

Basically there's no reason not to `yum install jq` in a tools container rather than relying on a statically linked binary.  One thing though at least I haven't really adapted into my workflow is using tools container binaries in pipes, but it definitely can be done.

miabbott:
@cgwalters Took your advice and went with a tools container...worked flawlessly in my tests  😄

cgwalters:
Cool, looks sane to me!  I'd like to make this pattern a lot nicer down the line..feels important enough to have a really first-class builtin flow.

miabbott:
> I need to make the `rpm_ostree_status` role use the version of `rpm-ostree` to determine when to filter the output through `jq`...

After running some of my own tests, I'm not sure this is necessary anymore.  If the "noisy keys" are not present in the JSON output, `jq` basically does a no-op an just returns all the JSON.

Removing the WIP label

mike-nguyen:
LGTM.  Nice work with the tools container.  It seems like the right workflow on atomic host.

Issue #299, opened 2017-12-19
migrate from "yes/no" values to real boolean "true/false" types


miabbott:
As part of #296, @ashcrow was quoting the `yes` and `no` values as that is proper YAML syntax.  Ansible is generous with their parsing and will automatically quote those values, so we were getting away with the unquoted usage.

To make things more compliant with the spec, we should change all the `yes/no` values to be real boolean `true/false` types.

mike-nguyen:
This issue was fixed in #302

PR #302, opened 2018-01-03
Comply to YAML spec where possible

mike-nguyen:
This commit fixes minor issues in the atomic host tests to comply
to the YAML spec.  The issues fixed were found using yamllint.

There were some instances of line length that could not be fixed
(i.e. Ansible when statements like  "'foo' in bar"). Using > or |
would break the logic so the line length had to be preserved.

### DIFF
--- a/tests/admin-unlock/main.yml
+++ b/tests/admin-unlock/main.yml
@@ -45,7 +45,7 @@
 
 - name: Ostree Admin Unlock - Setup
   hosts: all
-  become: yes
+  become: "yes"
 
   tags:
     - setup
@@ -78,7 +78,7 @@
 
 - name: Ostree Admin Unlock - Add/Remove Package - Non-persistence
   hosts: all
-  become: yes
+  become: "yes"
 
   tags:
     - unlock_add_remove_package_non_persistence
@@ -126,7 +126,7 @@
 
 - name: Ostree Admin Unlock - Hotfix - Add/Remove Package Persistence
   hosts: all
-  become: yes
+  become: "yes"
 
   tags:
     - hotfix_add_remove_package_persistence
@@ -188,7 +188,7 @@
 
 - name: Ostree Admin Unlock - Hotfix - Rollback
   hosts: all
-  become: yes
+  become: "yes"
 
   tags:
     - hotfix_rollback
@@ -213,7 +213,7 @@
 
 - name: Ostree Admin Unlock - No Install Without Unlock
   hosts: all
-  become: yes
+  become: "yes"
 
   tags:
     - cannot_install_without_unlock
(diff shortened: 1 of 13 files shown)


tests/admin-unlock/main.yml
```diff
@@ -45,7 +45,7 @@
 
 - name: Ostree Admin Unlock - Setup
   hosts: all
-  become: yes
+  become: "yes"
```
cgwalters: Weird, why?

mike-nguyen: yamllint complains -- warning  truthy value is not quoted  (truthy)

This is the rule: http://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.truthy

cgwalters: Ah and I see: http://yaml.org/type/bool.html

But something I don't quite understand here; the Ansible docs say the `become` flag is a boolean, so we shouldn't be converting it to a string AIUI.  

Or oh ugh, do we really need `!!bool true`?

miabbott: I think we can avoid the quoting by just using `become: true`

miabbott:
```
$ cat test.yml 
---
become: true
$ yamllint test.yml && echo $?
0
```

miabbott: This is in the same area as #299

mike-nguyen: I'll move it over to true/false

cgwalters: Ah true, looks like the same issue.

tests/pkg-layering/main.yml
```diff
@@ -235,21 +244,25 @@
...
+        pending_json['deployments'][0] is defined
+        and not pending_json['deployments'][0]['booted']
 
     - name: Set packages when deployment 1 is not booted
       set_fact:
         pending_pkgs: "{{ pending_json['deployments'][1]['packages'] }}"
-      when: pending_json['deployments'][1] is defined and not pending_json['deployments'][1]['booted']
+      when: >
+        pending_json['deployments'][1] is defined
+        and not pending_json['deployments'][1]['booted']
```
miabbott: My preference is to use a list for these kind of logical ANDs, but this is A-OK

```
when:
  - pending_json['deployments'][1] is defined
  - not pending_json['deployments'][1]['booted']
```

tests/pkg-layering/main.yml
```diff
@@ -235,21 +244,25 @@
...
 
-    - name: Fail if {{ g_pkg2 }} is in rpm-ostree status output for pending packages
+    - name: Fail if {{ g_pkg2 }} is in  pending packages
```
miabbott: nit: extra space between `in` and `pending`

miabbott:
A couple of minor nits, but this is overall OK.  I'd like to understand why the CI checks are failing because the failures don't look related to these changes.

mike-nguyen:
bot, retest this please

miabbott:
@mike-nguyen have you had a chance to look at these failures?

miabbott:
CentOS failures look like #316 

Fedora 26 failure due to `docker-latest` not being installed

Fedora 27 failure is infra flake


I don't see any huge functional changes here, so I'm inclined to merge this rather than fighting unrelated errors.

miabbott:
@mike-nguyen Feel free to merge based on comment ⬆️

mike-nguyen:
@miabbott I have all the tests running on RHELAH right now.  Once that passes I will merge.

mike-nguyen:
Merging this

PR #304, opened 2018-01-05
tree:  rework reboot role and usage

miabbott:
For far too long we have been using the `reboot` role by including the
`common/ans_reboot.yml` file.  Time to fix this weird design choice.

I've moved the `ans_reboot.yml` file to `roles/reboot/task/main.yml`
and updated all references to the old filename to use `include_role`.

While I was at it, I updated the `reboot` role to take a configurable
value for the `timeout`.  This was done after running our sanity test
against bare metal hosts that were either a) very old + slow or b) had
enormous amounts of RAM to initialize during a reboot.  (I'm talking
about observing reboot times approaching 10m!).

As part of the conversion to the new role, I've also deleted the
`ans_ah_head-1_deploy.yml` file because we no longer use this in the
tree.  If other folks are relying on this...well, we'll know about it
soon enough.  ¯\_(ツ)_/¯


mike-nguyen:
Makes sense and LGTM.  Thanks for the cleanup

PR #306, opened 2018-01-05
selinux_verify: adjust order of processes to be checked

miabbott:
When testing against systems that have slower CPUs or slower network
links, I was encountering a situation where I was not able to get the
SELinux label of the `rpm-ostree` daemon after using `rpm-ostree
status`.  I believe this was happening because the tests check the
SELinux label of `NetworkManager` first, allowing the `rpm-ostree`
process to spin down before we could check the label. (The daemon
appears to be activated via DBus (I think?), so it does not show
up in the running process list all the time.)

This change just puts `rpm-ostreed` at the top of the list so that it
is checked first.


mike-nguyen:
LGTM.

PR #307, opened 2018-01-05
test/i-s-t:  improve cleanup at end of test

miabbott:
This adds a role named `user_del` which will delete any users (with
prejudice!) passed as a list to the role.

The `improved-sanity-test` makes use of this new role at the very end
of the test and additionally calls `docker_remove_all` to cleanup any
docker artifacts (containers/images).

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -878,6 +878,15 @@
       tags:
         - rpm_ostree_cleanup_all
 
+    # remove any docker artifacts
+    - role: docker_remove_all
+
+    # remove users
+    - role: user_del
+      ud_users:
+        - "{{ g_uid1 }}"
+        - "{{ g_uid2 }}"
+
     # cleanup our subscriptions because we are nice people
     - when: ansible_distribution == 'RedHat'
       role: redhat_unsubscribe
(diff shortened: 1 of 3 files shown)


miabbott:
Marking as WIP because I honestly only did light testing and I'm not sure what is going to blow up (if anything)

miabbott:
Heh!  Passed all of our CI tests, so I'm sure it won't break anything  😏

tests/improved-sanity-test/main.yml
```diff
@@ -878,6 +878,15 @@
       tags:
         - rpm_ostree_cleanup_all
 
+    # remove any docker artifacts
+    - role: docker_remove_all
+
```
mike-nguyen: We seem to have tags for all our roles in this playbook.  Lets add them here to be consistent.

tests/improved-sanity-test/main.yml
```diff
@@ -878,6 +878,15 @@
...
+    # remove users
+    - role: user_del
+      ud_users:
+        - "{{ g_uid1 }}"
+        - "{{ g_uid2 }}"
+
```
mike-nguyen: Same here.

mike-nguyen:
Good to go.  Thanks Micah!

PR #310, opened 2018-01-09
README: new dashboard for 'next' tier tests

miabbott:
In the internal Red Hat infrastructure, we are now running a select
number of 'next' tier tests from the repo after a successful run of
the `improved-sanity-test`.  This commit creates a separate dashboard
that tracks the status of each of these tests.

The Markdown for this is hard to grok because of the long URLs that
are used all over the place, but the intent is to have a pass/fail
status badge for each test that will serve as a link to the result
log.


miabbott:
We should hold off on merging this until the internal jobs have the necessary changes to generate badges.

dustymabe:
is there a way to make it so that we can see which next tier tests passed and which ones failed?

mike-nguyen:
@dustymabe The badges that are generated by the internal job (pending) should indicate whether the test has passed or failed.

miabbott:
We have results for all the tests across all the streams, so I think this is ready for review or to be merged.

@mike-nguyen @dustymabe have a look over at my branch for the full render - https://github.com/miabbott/atomic-host-tests/tree/update_readme_nexttier

miabbott:
Might want to get #351 merged first...

dustymabe:
> @mike-nguyen @dustymabe have a look over at my branch for the full render - https://github.com/miabbott/atomic-host-tests/tree/update_readme_nexttier

I think I'm in love!

mike-nguyen:
Content is great, but the badges are a little small to read--especially the rawhide ones.  I wonder if we can reformat it somehow to get them bigger.

miabbott:
I spent an embarrassing amount of time trying all kinds of tricks to make those badges show up larger, but unfortunately I think the post-processing that GitHub does on the page effectively eliminates any of the sizing I tried.

Some alternatives to consider:
  - multiple 3x11 tables in the style of the 'sanity' test table...one table per 'next-tier' test
  - link from the README to a separate page with multiple tables as described above

Or we just live with the small badges.

dustymabe:
> Or we just live with the small badges.

+1

mike-nguyen:
Haha, I spent some time last night trying to get something to work too and came to the same conclusion.  Here's option 1 (i think)  that you suggested.  It ended up being very long.  
https://github.com/mike-nguyen/atomic-host-tests/blob/multiple_tables/README.md

I tried to put them side by side but the ended up not lining up nicely.
https://github.com/mike-nguyen/atomic-host-tests/blob/multiple_tables/README.md

mike-nguyen:
Another option, no header with two results side by side with an empty row between the next set of results.
https://github.com/mike-nguyen/atomic-host-tests/blob/multiple_tables/README3.md

miabbott:
@mike-nguyen I think I like `README2` the best, but `README3` is probably easier to maintain.  I'll see how it looks mashed into the main `README`

miabbott:
(I had a test branch I was iterating on, squashed my commmits, and cherry-picked the result to this branch.  Decided to force push the resulting commit)

I ended up going with the `README3` approach, just cleaned up some of the whitespace that was in @mike-nguyen's version.

You can see how it looks all put together - https://github.com/miabbott/atomic-host-tests/tree/update_readme_nexttier

The drawback to all these tables is that it pushes the remaining information on the README further down the page.

mike-nguyen:
Do you think we should move the status to another page? 

The only thing I would change in your iteration is to put the 2 fedora 26 streams next to each other and the fedora 27's underneath.

miabbott:
There's some value in leaving all the statuses up front; I think people are less likely to click through to another page.

Having said that, how about if we use an inline link to allow folks to click past the results to the additional information?

See what I've done here - https://github.com/miabbott/atomic-host-tests/tree/test_markdown

@dustymabe WDYT?

mike-nguyen:
I like the skip ahead link!  I think putting the 'Atomic Host Test' section up first so possible contributors :rofl: can see what the repo is about.  At the end of that section have a link to skip past the current test results to see more info.

miabbott:
Pushed a new version of the README at https://github.com/miabbott/atomic-host-tests/tree/test_markdown

I'm about done going through all these iterations.  :laughing: 

If anyone has any complaints, let the be known...otherwise I'm going to update this PR and get it merged.

mike-nguyen:
Nits, heading for Sanity Test Results and Nexttier should be smaller.  

We can probably add the F28 streams in at some point.  It doesn't need to be in this PR.

miabbott:
OK, changed the heading sizes.  Check the other branch.

I'll have a separate PR made for the F28 results after this gets merged.

mike-nguyen:
LGTM

miabbott:
Rebased, cherry-picked the commit from the other branch, and force pushed :arrow_up:

mike-nguyen:
Nice work!

PR #311, opened 2018-01-10
Begin block statements with when for all tests.

kbidarkar:


kbidarkar:
My bad, looks like I submitted too early, will go through and update as per the contribution guide.
But curious, why the checks have failed, am I missing something ?

mike-nguyen:
@kbidarkar Looks like mostly infra related failures.  I'll look a little deeper.

miabbott:
@kbidarkar You've just experienced the flakiness of our test infra 😢 

`Failed to connect to the host via ssh: mux_client_hello_exchange: write packet: Broken pipe`

That error points to the networking stack of the VM or cloud (we use an internal OpenStack instance) went down.  Unfortunately, we see this more often than we like.


`409 Client Error: Conflict ("conflict: unable to delete da201d6f58c0 (must be forced) - image is being used by stopped container 42c5823997de")`

This error looks like we need to modify the `atomic` test suite to remove any existing containers before using the `atomic_images_delete_all`  role

miabbott:
Straightforward changes; I'd suggest we try to resolve the error in the `atomic` test suite in a separate PR, then merge this in.

mike-nguyen:
I'm looking at the `atomic` test suite right now but got a different error around deleting images.

miabbott:
@kbidarkar we're experiencing infrastructure problems (again!) for our CI tests; I'll restart these checks once the issues are resolved

miabbott:
@rh-atomic-bot bot, retest this please

miabbott:
bot, retest this please

tests/docker/main.yml
```diff
@@ -114,6 +114,7 @@
 
   pre_tasks:
     - block:
+      when: g_docker_latest
```
mike-nguyen: This was a failure in the log.  I'm not sure how this syntax will work here.  I would try
```
- when: g_docker_latest
  block:
    - name: ...
```

miabbott:
The CI tests are a mess, but none of the failures are due to changes here.  I've been able to run the tests appropriately locally. Going to merge this in.

PR #312, opened 2018-01-10
Use force option for atomic images delete all

mike-nguyen:
The atomic_images_delete_all role was failing because the atomic
images delete command cannot delete an image if it is used in
an exited container.  The --force option is used to work around
this.


miabbott:
LGTM

PR #314, opened 2018-01-15
Install k8 worker nodes and etcd system containers for Fedora 27AH

kbidarkar:
This is required, as from Fedora27AH, k8, flannel and etcd are no
longer part of the Host's image.

### DIFF
--- a/roles/k8_cluster_services_rc_setup/tasks/main.yml
+++ b/roles/k8_cluster_services_rc_setup/tasks/main.yml
@@ -25,45 +25,57 @@
 
   - name: insert private registry ip in webserver controller image yml
     replace:
-      dest=/root/db-rc.yml
-      regexp='PRIVATE'
-      replace='{{ ansible_docker0.ipv4.address }}'
+      dest: /root/db-rc.yml
+      regexp: 'PRIVATE'
+      replace: '{{ ansible_docker0.ipv4.address }}'
 
   - name: insert private registry ip in db controller image yml
     replace:
-      dest=/root/webserver-rc.yml
-      regexp='PRIVATE'
-      replace='{{ ansible_docker0.ipv4.address }}'
+      dest: /root/webserver-rc.yml
+      regexp: 'PRIVATE'
+      replace: '{{ ansible_docker0.ipv4.address }}'
 
   - name: wait for kubernetes apiserver to be available
     wait_for:
       port: 8080
       timeout: 120
 
+  - name: Define kubectl path for F27AH
+    when: (ansible_distribution == "Fedora" and ansible_distribution_major_version > '26') or
+          (ansible_distribution == "CentOSDev")
+    set_fact:
+      kubectl_path: "/usr/local/bin/kubectl"
+
+  - name: Define kubectl path for F26AH, Centos/RHEL
+    when: (ansible_distribution == "RedHat") or
+          (ansible_distribution == "Fedora" and ansible_distribution_major_version < '27') or
+          (ansible_distribution == "CentOS")
+    set_fact:
+      kubectl_path: "/usr/bin/kubectl"
 
   - name: create db and webserver service
-    command: kubectl create -f /root/{{ item }}
+    command: "{{ kubectl_path }} create -f /root/{{ item }}"
     with_items:
       - db-service.yml
       - webserver-service.yml
     ignore_errors: True
 
   - name: create db and webserver replication controller
-    command: kubectl create -f /root/{{ item }}
+    command: "{{ kubectl_path }} create -f /root/{{ item }}"
     with_items:
       - db-rc.yml
       - webserver-rc.yml
     ignore_errors: True
 
   - name: verify db pod is running
-    shell: kubectl get pods | grep db-controller
+    shell: "{{ kubectl_path }} get pods | grep db-controller"
     register: output
     until: output.stdout.find("Running") > -1
     retries: 12
     delay: 10
 
   - name: verify webserver-controller is running
-    shell: kubectl get pods | grep webserver-controller
+    shell: "{{ kubectl_path }} get pods | grep webserver-controller"
     register: output
     until: output.stdout.find("Running") > -1
     retries: 12
--- a/roles/kubernetes_setup/tasks/main.yml
+++ b/roles/kubernetes_setup/tasks/main.yml
@@ -9,7 +9,9 @@
     rv_rpms: kubernetes-node
 
 - name: Define kube 1.5 vars
-  when: g_atomic_host['kubernetes-node'] | version_compare('1.6', '<')
+  when:
+    - g_atomic_host is defined
+    - g_atomic_host['kubernetes-node'] | version_compare('1.6', '<')
   set_fact:
     kube_maj_ver: "1.5"
     kube_apiserver: "registry.access.redhat.com/rhel7/kubernetes-apiserver"
@@ -17,14 +19,31 @@
     kube_scheduler: "registry.access.redhat.com/rhel7/kubernetes-scheduler"
 
 - name: Define kube 1.6 images
-  when: g_atomic_host['kubernetes-node'] | version_compare('1.6', '>=')
+  when:
+    - g_atomic_host is defined
+    - g_atomic_host['kubernetes-node'] | version_compare('1.6', '>=')
   set_fact:
     kube_maj_ver: "1.6"
     kube_apiserver: "registry.fedoraproject.org/f26/kubernetes-apiserver"
     kube_controller_manager: "registry.fedoraproject.org/f26/kubernetes-controller-manager"
     kube_scheduler: "registry.fedoraproject.org/f26/kubernetes-scheduler"
 
+- name: Define kube 1.7 master nodes, worker nodes and etcd images
+  when: (ansible_distribution == "Fedora" and ansible_distribution_major_version > '26') or
+        (ansible_distribution == "CentOSDev")
+
+  set_fact:
+    kube_maj_ver: "1.7"
+    kube_apiserver: "registry.fedoraproject.org/f27/kubernetes-apiserver"
+    kube_controller_manager: "registry.fedoraproject.org/f27/kubernetes-controller-manager"
+    kube_scheduler: "registry.fedoraproject.org/f27/kubernetes-scheduler"
+    kube_kubelet: "registry.fedoraproject.org/f27/kubernetes-kubelet"
+    kube_proxy: "registry.fedoraproject.org/f27/kubernetes-proxy"
+    etcd: "registry.fedoraproject.org/f27/etcd"
+
 - name: pull kubernetes api-server, controller-mgr, scheduler
+  when:
+    - g_atomic_host is defined
   command: docker pull {{ item }}
   register: pull
   failed_when: item not in pull.stdout
@@ -32,6 +51,37 @@
     - "{{ kube_apiserver }}"
     - "{{ kube_controller_manager }}"
     - "{{ kube_scheduler }}"
+
+- block:
+    - name: Pull k8 and etcd images from the registry
+      command:  atomic pull --storage ostree "{{ item }}"
+      register: ap
+      retries: 3
+      delay: 10
+      until: "ap.rc == 0"
+      with_items:
+        - "{{ kube_apiserver }}"
+        - "{{ kube_controller_manager }}"
+        - "{{ kube_scheduler }}"
+        - "{{ kube_kubelet }}"
+        - "{{ kube_proxy }}"
+        - "{{ etcd }}"
+
+    - name: Install master node system containers that is, k8-apiserver, k8-controller_manager and k8-scheduler
+      command: "{{ item }}"
+      with_items:
+        - atomic install --system --system-package=no --name kube-apiserver {{ kube_apiserver }}
+        - atomic install --system --system-package=no --name kube-controller-manager {{ kube_controller_manager }}
+        - atomic install --system --system-package=no --name kube-scheduler {{ kube_scheduler }}
+
+    - name: Install worker node system containers, that is k8-kubelet, k8-proxy and etcd
+      command: "{{ item }}"
+      with_items:
+        - atomic install --system --system-package=no --name kubelet {{ kube_kubelet }}
+        - atomic install --system --system-package=no --name kube-proxy {{ kube_proxy }}
+        - atomic install --system --system-package=no --storage=ostree --name etcd {{ etcd }}
+  when: (ansible_distribution == "Fedora" and ansible_distribution_major_version > '26') or
+        (ansible_distribution == "CentOSDev")
 
 - name: make db directory
   file:
--- a/roles/rpm_version/tasks/version.yml
+++ b/roles/rpm_version/tasks/version.yml
@@ -15,8 +15,10 @@
 - name: Determine version of {{ rv_rpm }}
   command: rpm -q --queryformat '%{VERSION}' {{ rv_rpm }}
   register: rpmq
+  ignore_errors: true
 
 - name: Set facts
+  when: rpmq.rc == 0
   set_fact:
     rv_version: "{{ rpmq.stdout }}"
     g_atomic_host: "{{ g_atomic_host|default({}) | combine( { rv_rpm: rpmq.stdout } ) }}"
--- a/roles/k8_remove_all/tasks/main.yml
+++ b/roles/k8_remove_all/tasks/main.yml
@@ -17,11 +17,25 @@
     - kube-proxy
     - kubelet
 
+- name: Define kubectl path for F27AH
+  when: (ansible_distribution == "Fedora" and ansible_distribution_major_version > '26') or
+        (ansible_distribution == "CentOSDev")
+  set_fact:
+    kubectl_path: "/usr/local/bin/kubectl"
+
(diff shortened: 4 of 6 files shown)


roles/kubernetes_setup/tasks/main.yml
```diff
@@ -33,6 +33,35 @@
...
+    kube_proxy: "registry.fedoraproject.org/f27/kubernetes-proxy"
+    etcd: "registry.fedoraproject.org/f27/etcd"
+
+
+- name: Install master node system containers that is, k8-apiserver, k8-controller_manager and k8-scheduler
+  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
+  command: atomic install --system --system-package=no --name kube-apiserver {{ kube_apiserver }}
+  command: atomic install --system --system-package=no --name kube-controller-manager {{ kube_controller_manager }}
+  command: atomic install --system --system-package=no --name kube-scheduler {{ kube_scheduler }}
+
```
mike-nguyen: I think these commands have to be in separate tasks or put into a loop.  Only one command will run (at least that is how it is on Ansible 2.2.x).

```
- name: Install master node system containers that is, k8-apiserver, k8-controller_manager and k8-scheduler
  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
￼  command: {{ item }}
  with_items:
    - atomic install --system --system-package=no --name kube-apiserver {{ kube_apiserver }}
￼    - atomic install --system --system-package=no --name kube-controller-manager {{ kube_controller_manager }}
￼    - atomic install --system --system-package=no --name kube-scheduler {{ kube_scheduler }}
```

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -33,6 +33,35 @@
...
+- name: Install worker node system containers, that is k8-kubelet, k8-proxy and etcd
+  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
+  command: atomic install --system --system-package=no --name kubelet {{ kube_kubelet }}
+  command: atomic install --system --system-package=no --name kube-proxy {{ kube_proxy }}
+  command: atomic install --system --system-package=no --storage=ostree --name etcd {{ etcd }}
+
```
mike-nguyen: Same as above

roles/kubernetes_setup/tasks/main.yml
kbidarkar: @mike-nguyen  Thanks, I was so much looking at k8 and other things, I missed this totally.

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -33,6 +33,39 @@
     - "{{ kube_controller_manager }}"
     - "{{ kube_scheduler }}"
 
+- name: Define kube 1.7 master nodes images
+  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
```
miabbott: nit:  I prefer the list format for logical ands:

```
when:
  - ansible_distribution == "Fedora"
  - ansible_distribution_major_version > '26'
```

miabbott: For version comparisons, it is often better to use the `version_compare` filter, but in this case it should be fine as is.

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -33,6 +33,39 @@
...
+    - atomic install --system --system-package=no --name kube-scheduler {{ kube_scheduler }}
+
+- name: Install worker node system containers, that is k8-kubelet, k8-proxy and etcd
+  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
+  command: {{ item }}
+  with_items:
+    - atomic install --system --system-package=no --name kubelet {{ kube_kubelet }}
+    - atomic install --system --system-package=no --name kube-proxy {{ kube_proxy }}
+    - atomic install --system --system-package=no --storage=ostree --name etcd {{ etcd }}
+
```
miabbott: All these changes could be contained in a `block:` statement, so that we are only doing the `when:` check a single time.

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -33,6 +33,39 @@
...
+    kube_scheduler: "registry.fedoraproject.org/f27/kubernetes-scheduler"
+
+- name: Define kube 1.7 worker nodes and etcd images
+  when: ansible_distribution == "Fedora" and ansible_distribution_major_version > '26'
+  set_fact:
+    kube_maj_ver: "1.7"
+    kube_kubelet: "registry.fedoraproject.org/f27/kubernetes-kubelet"
+    kube_proxy: "registry.fedoraproject.org/f27/kubernetes-proxy"
+    etcd: "registry.fedoraproject.org/f27/etcd"
+
```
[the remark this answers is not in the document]
kbidarkar: I thought of this and wrote it this way initially, but then realized the `kubernetes-node` pkg does not exist on f27AH and may never set the values for f27AH, hence changed it. 
Is there a reason we are checking specifically for `kubernetes-node` pkg, can I make it more generic like, just checking it for fedora and 26.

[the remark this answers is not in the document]
kbidarkar: Ah!, I know, we are also checking for Centos and RHEL I believe and hence so.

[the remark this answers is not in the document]
miabbott: Hmm...the exclusion of `kubernetes-node` on F27 might require additional rework to this role.  I'm not certain what happens if the RPM that is being checked via `rpm_version` is not there on the host.  

Maybe we need to add a `when:` to the use of the `rpm_version` so that it only runs on CentOS + RHEL?

@mike-nguyen any thoughts here?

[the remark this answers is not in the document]
mike-nguyen: I just checked the `rpm_version` role and it will set the version to the standard output of `rpm -q --queryformat '%{VERSION}' rpm_name`.  When the package is not installed it returns `package rpm_name is not installed` which is not ideal.  

Maybe we should modify `rpm_version` with a when statement to only set_facty if 'is not installed' is not in standard output.  Then for the comparison, we can do the comparison AND `g_atomic_host['kubernetes-node'] is defined`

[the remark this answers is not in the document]
miabbott: Seems like a good approach; I would just rely on the return code rather than the stderr.  For example:

```
$ rpm -q --queryformat '%{VERSION}' foo
package foo is not installed
$ echo $?
1
$ rpm -q --queryformat '%{VERSION}\n' kernel
4.14.11
$ echo $?
0
```

As long as the RC == 0, we can save off the version.  @mike-nguyen can you make a PR for this change?

miabbott:
bot, retest this please

roles/rpm_version/tasks/version.yml
```diff
@@ -17,6 +17,7 @@
   register: rpmq
```
mike-nguyen: `ignore_errors: true`

roles/rpm_version/tasks/version.yml
```diff
@@ -17,6 +17,7 @@
...
 
 - name: Set facts
+  when: "'is not installed' not in rpmq.stdout"
```
mike-nguyen: `when: rpmq.rc == 0`

kbidarkar:
what is the difference between `centos/7/atomic` and `centos/7/atomic/continuous` ?

jlebon:
@kbidarkar Check out PAPR's [DISTROS.md](https://github.com/projectatomic/papr/blob/master/docs/DISTROS.md#centos-streams). Essentially, the `/continuous` version uses git master versions of Project Atomic apps like atomic, ostree, rpm-ostree, and others.

kbidarkar:
For `fedora/27/atomic` : https://s3.amazonaws.com/aos-ci/ghprb/projectatomic/atomic-host-tests/2edccc4aac7afc8de7a436feafda1263d3dd97d9.0.1516741900478191583/output.log

Looks like this is the issue, https://github.com/kubernetes/kubectl/issues/30

kbidarkar:
Was looking today at f26 and f27 issues in particular.

ansible-playbook -vi inventory --private-key=/home/kbidarka/atomic_kbidarka.pem tests/k8-cluster/main.yml -u fedora

For F26:
-----------
First Installed F26 AH and then upgraded the setup, I had to do this as kubernetes-1.53 was on F26.

F26 failures: 
papr run: https://s3.amazonaws.com/aos-ci/ghprb/projectatomic/atomic-host-tests/2edccc4aac7afc8de7a436feafda1263d3dd97d9.0.1516741900478191583/output.log
The issue could be : https://github.com/kubernetes/kubectl/issues/30

For F27:
----------

First installed F26 AH and then upgraded to F27 by following, https://www.projectatomic.io/blog/2017/11/fedora-atomic-26-to-27-upgrade/

F27 failures: 
local run: https://paste.fedoraproject.org/paste/lGsMzVAFmU0043~l1kyz-g
papr run: https://s3.amazonaws.com/aos-ci/ghprb/projectatomic/atomic-host-tests/2edccc4aac7afc8de7a436feafda1263d3dd97d9.1.1516741909020425976/output.log

The issue : For some reason when running the below command,  
a) "kubectl create -f /root/db-service.yml", ansible throws error saying "[Errno 2] No such file or directory"
b) "kubectl get pods | grep db-controller", ansible throws error saying "/bin/sh: kubectl: command not found"
NOTE: As seen from the logs, ` registry.fedoraproject.org/f27/kubernetes-apiserver` system container does bring in `kubectl` as when logged in and checked on the f27AH machine.

The errors for F27AH appear to be weird, any idea what could be the issue ?

Will look at `centos/7/atomic/continuous` tomorrow.

mike-nguyen:
Micah andI looked through the failures and these are our findings.

**F26**
I just submitted PR #323 that should fix the issue you are seeing.

**F27**
The kube-apiserver system container installs kubectl in /usr/local/bin/kubectl.  Sudo apparently starts with a minimal environment that does not include /usr/local/bin (see more at http://ask.xmodulo.com/define-path-environment-variable-sudo-commands.html).  To work around this, there are two options.

1. Edit the /etc/sudoers files and add the path as the link above suggests and then remove that path at the end of the test.
2. Set the full path to the kubectl command (/usr/loca/bin/kubectl) as a fact and use it whenever you run the command.  The kubectl command fact will need to be conditionalized for the containerized kubernetes install case, and the kubernetes installed as an rpm case.  When installed as an rpm kubectl will be in /usr/bin/kubectl or called directly using kubectl.

The preferred choice is option 2, but if  for some reason that won't work, use option 1.

kbidarkar:
Will add in the changes required from #323, will remove it once I rebase it.

kbidarkar:
I thought `kubernetes-node` package to exist on centos-7 for `centos/7/atomic/continuous` to work.
Maybe I should treat the `centos/7/atomic/continuous`  too as f27AH for k8-cluster tests ?

kbidarkar:
For F27AH, we now see the below issue:
```
STDERR:
---
time="2018-01-29T12:33:51Z" level=fatal msg="Error writing blob: unexpected EOF"
```
The previous issues have been fixed.

kbidarkar:
Looks like shortage of storage space on `centos/7/atomic/continuous` when installing k8 system containers: 
```
 STDERR:
---
time="2018-01-30T04:24:34Z" level=fatal msg="Error committing the finished image: /builddir/build/BUILD/skopeo-2e8377a7085c13674be156f32c100708da838b7e/vendor/src/github.com/ostreedev/ostree-go/pkg/otbuiltin/commit.go:407 - Writing content object: min-free-space-percent '3%' would be exceeded, 109.8 MB more required" 
---
```

miabbott:
The network issues are a pain to deal with; the infra hasn't been cooperating lately.

Maybe we should try pulling the images first before trying to install them?  We can implement some retry logic to handle the pull timing out.

Something like this, but with the right image names:

```yaml
- name: pull images
  command:  atomic pull --storage ostree {{ item }}
  register: ap
  retry: 5
  delay: 60
  until: ap.rc | success
  with_items:
    - kube-apiserver
    - kube-controller-mgr
```

miabbott:
I don't know why the `centos/7/atomic/continuous` case would run out of storage...I'll have to dig into that further

miabbott:
@kbidarkar I was able to reproduce the out of storage error on `centos/7/atomic/continuous`.  I think the best option is to try to resize the LV before pulling down any of the system containers.

I've got a PR in #326 to add a new role to assist with the resizing.  Feel free to review over there and then you could incorporate it into this PR.

You should probably rebase this PR as it is, to get changes like #323 and #319 which will affect the `k8-cluster` test.

roles/k8_cluster_services_rc_setup/tasks/main.yml
```diff
@@ -25,45 +25,58 @@
...
 
   - name: wait for kubernetes apiserver to be available
     wait_for:
       port: 8080
       timeout: 120
 
+  - name: Define kubectl path for F27AH
+    when:
+      - (ansible_distribution == "Fedora" and ansible_distribution_major_version > '26')
+      - (ansible_distribution == "CentOSDev")
```
miabbott: This should be a logical `OR`

kbidarkar: My bad will fix this.

kbidarkar:
After #326 is merged, plan to handle its changes during the rebase. Just added it here for the tests to pass.

miabbott:
@kbidarkar I ran this against F27AH myself and hit similar problems.

I ended up like this:

```
# kubectl get all
2018-02-01 20:54:20.997546 I | proto: duplicate proto type registered: google.protobuf.Any
2018-02-01 20:54:20.997711 I | proto: duplicate proto type registered: google.protobuf.Duration
2018-02-01 20:54:20.997767 I | proto: duplicate proto type registered: google.protobuf.Timestamp
NAME                                   READY     STATUS             RESTARTS   AGE
po/kube-apiserver-127.0.0.1            1/1       Running            0          14m
po/kube-controller-manager-127.0.0.1   0/1       CrashLoopBackOff   7          15m
po/kube-scheduler-127.0.0.1            0/1       CrashLoopBackOff   7          15m

NAME                      DESIRED   CURRENT   READY     AGE
rc/db-controller          1         0         0         15m
rc/webserver-controller   1         0         0         15m

NAME                    CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
svc/db-service          10.254.191.247   <none>        3306/TCP   15m
svc/kubernetes          10.254.0.1       <none>        443/TCP    15m
svc/webserver-service   10.254.245.26    <none>        80/TCP     15m
```

And looking at the logs for the failed pods, both have similar errors:

```
# kubectl logs po/kube-scheduler-127.0.0.1                           
2018-02-01 20:52:40.595048 I | proto: duplicate proto type registered: google.protobuf.Any               
2018-02-01 20:52:40.595229 I | proto: duplicate proto type registered: google.protobuf.Duration          
2018-02-01 20:52:40.595295 I | proto: duplicate proto type registered: google.protobuf.Timestamp         
/usr/bin/kube-scheduler-docker.sh: line 3: /etc/kubernetes/scheduler: Permission denied                  
/usr/bin/kube-scheduler-docker.sh: line 4: /etc/kubernetes/config: Permission denied                     
2018-02-01 20:49:38.765531 I | proto: duplicate proto type registered: google.protobuf.Any               
2018-02-01 20:49:38.766373 I | proto: duplicate proto type registered: google.protobuf.Duration          
2018-02-01 20:49:38.766730 I | proto: duplicate proto type registered: google.protobuf.Timestamp         
W0201 20:49:38.822553     113 client_config.go:529] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.                                                             
W0201 20:49:38.823214     113 client_config.go:534] error creating inClusterConfig, falling back to default config: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory            
Error: unable to create kube client: unable to build config from flags: invalid configuration: no configuration has been provided                                                                                  
Error: unable to create kube client: unable to build config from flags: invalid configuration: no configuration has been provided                                            
```

Looks like a permissions error trying to read the configuration files in `/etc/kubernetes` but on my host the permissions look sane.

@mike-nguyen any ideas?

I'll dig around some more to see if I can get it sorted out.

miabbott:
I think I figured out what needs to be done.  It'll require some significant rework of the `kubernetes_setup` role because of some slight changes that apply only for the system containers.

I've detailed the steps I went through in this gist - https://gist.github.com/miabbott/549655e4557c7bd9f774f8d45a79cac6

@kbidarkar Let me know if you will continue to work on this or if you would rather we take over from here.

kbidarkar:
@miabbott, Thanks for the gist file, will take a look at the changes needed and let me try to update it quickly, can work on this till Friday and then will be on PTO for 2 weeks. Would be much interested in getting this PR to work. :)

miabbott:
I spoke with @kbidarkar recently and he will be on PTO for the next two weeks.  In order to get these changes into the repo sooner rather than later, I've taken his commits and made a new PR (in #335) that finishes the work he started.  I'm going to close this PR in favor of #335.

Many thanks to Kedar for taking this on despite not having any prior experience with kubernetes!

mike-nguyen:
Thanks for the contribution @kbidarkar.  It is not easy getting all this stuff working on all the different streams!

Issue #317, opened 2018-01-17
modify part of 'admin-unlock' test to not retry install of RPM


miabbott:
The "No Install Without Unlock" portion of the `admin-unlock` test suite attempts to install an RPM before using `ostree admin unlock` as a negative test.  Continually retrying this install is fruitless because we expect it to fail.  The behavior should be changed to only try the install once.

```
PLAY [Ostree Admin Unlock - No Install Without Unlock] *************************

TASK [setup] *******************************************************************
ok: [testnode]

TASK [Attempt to install rpm] **************************************************
FAILED - RETRYING: TASK: Attempt to install rpm (5 retries left).
FAILED - RETRYING: TASK: Attempt to install rpm (4 retries left).
FAILED - RETRYING: TASK: Attempt to install rpm (3 retries left).
FAILED - RETRYING: TASK: Attempt to install rpm (2 retries left).
FAILED - RETRYING: TASK: Attempt to install rpm (1 retries left).
fatal: [testnode]: UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}
```

miabbott:
Closed via #318

PR #318, opened 2018-01-17
Don't retry when expecting failures

mike-nguyen:
Addresses #317 .  Retries were added to increase reliability of tests
when performing network operations, however, we should not retry
when we expect a failure.

I also added a minor fix to only fail when the command succeeds.


miabbott:
This is a simple enough fix, I'm not going to fret about our infra flakes.  🙄

PR #321, opened 2018-01-18
docker_build_httpd: cleanup file layout

miabbott:
In anticipation of having a generic `docker_build` role, I moved some
files around and made some slight tweaks to the `docker_build_httpd`
role.

A handy change included here is putting a dependency on
`osname_set_fact`, so that `docker_build_httpd` should always have the
necessary variables defined.


mike-nguyen:
Does this have to be merged after #330?  I got some failures building the images on CentOS and Fedora.

miabbott:
Hmm...it might be best to get #330 in first, then deal with whatever errors you are hitting here.

mike-nguyen:
bot, retest this please

mike-nguyen:
Cleaned up nicely after merging #330!

PR #322, opened 2018-01-19
journal_fatal_msgs: add check for SELinux denials from kernel

miabbott:
In RHBZ#1536690, I observed there were SELinux denials present on the
host, but our tests were catching them.  These denial messages were on
the 'kernel' transport and we were only checking the 'audit' transport,
so I've added a new rule to look for denials from that 'kernel'
transport.


miabbott:
The side-effect of these changes is that it is going to cause the RHEL + CentOS AH streams to fail until the above mentioned RHBZ is fixed.  (Also the reason the `centos/7/atomic` stream failed)

mike-nguyen:
LGTM!

PR #323, opened 2018-01-25
Fix action script

mike-nguyen:
Closing the mysqldb connection inside the with statement block was
causing segmentation faults and tracebacks in newer versions of
python.  This caused errors getting the action page from Apache.

Moving the connection close outside the with statement fixed this
on fedora, rhel, and centos.


miabbott:
LGTM

PR #326, opened 2018-01-30
roles: new resize_lv role

miabbott:
We may want to resize a logical volume during our tests, in order to
make room for `ostree` host content or system container content.  This
role provides simple functionality to do so.

### DIFF
--- a/roles/resize_lv/tasks/main.yml
+++ b/roles/resize_lv/tasks/main.yml
@@ -7,21 +7,19 @@
 # Atomic Host systems, the logical volume names are well known, but the
 # volume group names can differ based on install method.
 #
-# If you have advance knowledge of the sizes of the volume groups and
-# logical volumes, you can specify the desired size of the logical volume,
-# otherwise it defaults to consuming 50% of the available space in the
-# volume group.
+# Additionally, you must specify the desired size of the logical volume
+# in units of G.  If the current size of the logical volume is less than
+# the desired value, the volume will be extended.
 #
 # Required variables:
-#  - rl_lvname:  name of the logical volume
-#
-# Optional variables:
+#   - rl_lvname:  name of the logical volume
 #   - rl_lvsize:  desired size in G of the logical volume
 
 - name: Fail if variables aren't set
-  when: rl_lvname is not defined
+  when: rl_lvname is not defined or
+        rl_lvsize is not defined
   fail:
-    msg: "The variable 'rl_lvname' was not defined"
+    msg: "The variables 'rl_lvaname' and 'rl_lvsize' must be defined"
 
 - name: Determine the name of the volume group
   command: vgs --no-headings -o vg_name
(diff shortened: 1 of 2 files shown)


cgwalters:
This isn't idempotent though, right?  I think it might work better to just "test if root LV is <= 5G" first...I swear we've written a playbook like that somewhere.

miabbott:
> This isn't idempotent though, right? I think it might work better to just "test if root LV is <= 5G" first...I swear we've written a playbook like that somewhere.

Yeah, that is a good point.  I'll update with that kind of logic.

miabbott:
Hmm...most of the time, I think this is going to be used to extend the `root` LV, but to make it more flexible I've allowed the user to specify the LV name in case we have a need for that.  Checking just the size of the `root` LV won't fly in that case.

I guess I can just require the user to provide a desired size of the LV and check to see if the current size is at least that big.  That way the LV won't just keep growing by the `50%FREE` if this is used multiple times.

kbidarkar:
I have added this patch manually and testing it now on my branch.
I have the below in `/tests/k8-cluster/vars.yml`
```
rl_lvname: 'root'
```

kbidarkar:
For some reason, this does not work. so changed it this way, https://github.com/projectatomic/atomic-host-tests/pull/314/files#diff-cb08536f06ae39534b85bad3675672e5

miabbott:
@kbidarkar I used the updated commit on this PR on top of your PR and I got most of the way through the `k8-cluster` playbook on CAHC

Re-running the `resize_lv` role a few times is now idempotent, since we specify the size and check to see if it is at least that size or larger.

miabbott:
@kbidarkar 

```diff
$ git diff tests/k8-cluster/main.yml 
diff --git a/tests/k8-cluster/main.yml b/tests/k8-cluster/main.yml
index 6502125..82fd55c 100644
--- a/tests/k8-cluster/main.yml
+++ b/tests/k8-cluster/main.yml
@@ -37,6 +37,13 @@
       tags:
         - redhat_subscription
 
+    - when: ansible_distribution == 'CentOSDev'
+      role: resize_lv
+      rl_lvname: 'root'
+      rl_lvsize: 6
+      tag:
+          - resise_lv
+
     - role: docker_private_registry
       tags:
         - docker_private_registry
```

kbidarkar:
I tested with this PR included and realized we need to also `resize` it, when looking at the logs, https://s3.amazonaws.com/aos-ci/ghprb/projectatomic/atomic-host-tests/05ce188ddacbcceb6fd96b5cd2d9dfb6824b3473.3.1517445636554524087/output.log

For which I added the below stuff and it seems to work now. Let me run the `k8-cluster` playbook now.

```
+- name: Grow the logical volume size
+  when: rl_lvsize_current|int < rl_lvsize|int
+  filesystem:
+    fstype: xfs
+    dev: "/dev/{{ rl_vgname }}/{{ rl_lvname }}"
+    resizefs: yes
```

roles/resize_lv/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+  set_fact:
+    rl_vgname: "{{ rl_vgs_name.stdout|trim }}"
+
+- name: Get current size of logical volume
+  command: "lvs -o lv_size --no-headings --units G --nosuffix {{ rl_vgname }}/{{ rl_lvname }}"
+  register: rl_lvs_size
+
+- name: Set current lvsize fact
+  set_fact:
+    rl_lvsize_current: "{{ rl_lvs_size.stdout|trim|round|int }}"
```
kbidarkar: I also had to update this one to `rl_lvsize_current: "{{ rl_lvs_size.stdout|trim|float|round|int }}"`

miabbott:
Thanks for the comments @kbidarkar!  I've made additional changes in the most recent commit

miabbott:
These CI checks are pretty useless since this a new role and won't actually get tested as part of the checks.

I promise it works for me, though!  :)

roles/resize_lv/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Required variables:
+#   - rl_lvname:  name of the logical volume
+#   - rl_lvsize:  desired size in G of the logical volume
+
+- name: Fail if variables aren't set
+  when: rl_lvname is not defined or
+        rl_lvsize is not defined
+  fail:
+    msg: "The variables 'rl_lvaname' and 'rl_lvsize' must be defined"
```
mike-nguyen: nit: misspelled rl_lvname

kbidarkar:
I feel we can now merge this PR. I will rebase my PR once this is merged and get rid of same stuff from this PR while rebasing.

miabbott:
Thanks for the review, folks!  I'm going to merge this.

PR #327, opened 2018-02-02
bdsf: utilize the tools container to hide noisy JSON keys

miabbott:
Similar to the `rpm_ostree_status` role, when we output the status in
JSON format, one of the keys has a list of all the packages that are
included in the ostree commit.  In f6a7086, we introduced the use of a
tools container to pipe the JSON output through `jq` to hide the noisy
keys.  This is basically the same implementation.

(We probably should consolidate the two roles, but that is for another
PR...)

### DIFF
--- a/roles/booted_deployment_set_fact/meta/main.yml
+++ b/roles/booted_deployment_set_fact/meta/main.yml
@@ -1,2 +1,6 @@
 ---
 allow_duplicates: true
+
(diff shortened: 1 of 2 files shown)


miabbott:
The `centos/7/atomic` failures are related to #322

mike-nguyen:
LGTM

PR #329, opened 2018-02-07
journal_fatal_msgs: add tags for each fatal message

miabbott:
We have no way to skip the check for the individual 'fatal' messages
that are covered by the `journal_fatal_msgs` role.  By adding tags to
each message we are looking for, we can selectively disable the check
for individual messages that we know still exist and allow us to
execute the remaining playbook.


mike-nguyen:
LGTM

PR #330, opened 2018-02-07
docker_build_httpd: refresh Fedora images used

miabbott:
The test was still building images using Fedora 24/25 and those are
pretty old, so remove them and start building with Fedora 26/27.

PR #331, opened 2018-02-08
roles: convert docker roles to be more generic

miabbott:
We have a number of specialized roles that are used in the build, run,
and removal of an `httpd` container.  While this has worked thusfar,
it limits the flexibility around the build/run/remove operations of
Docker containers.

This change modifies the existing roles to be more generic and allows
us to use them with more flexibility.

As part of the changes, a new role named `check_open_port` was added
to replace some functionality that was removed from an exiting role.

Finally, the existing tests that used the roles were updated to use
the new role names.


miabbott:
This PR includes the commit from #321 and also requires #330 to be merged.

So when those two PRs get in, I'll rebase this PR and we can start proper review again.

miabbott:
I've rebased this and force pushed the result here.  Time for reviews!

roles/docker_build/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
+# vim: set ft=ansible:
+#
+# This is a generic role to handle the building of docker images when a
+# Dockerfile is provided.  You need to provide the location of the
+# Dockerfile (presumably in the repo) and the name you wish to call your
+# newly built image.
+#
+# NOTE:  the role uses the `syncrhonize` module to copy files to the host
```
mike-nguyen: s/syncrhonize/synchronize

roles/docker_build/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Set build_dir fact
+  set_fact:
+    build_dir: "{{ m.stdout }}"
+
+- name: Copy files to host
+  synchronize:
+    src: "{{ db_src }}"
+    dest: "{{ build_dir }}"
+
+- name: Build httpd image
```
mike-nguyen: s/httpd/docker?

roles/docker_build/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  register: db
+  retries: 5
+  delay: 60
+  until: db|success
+
+- name: Get docker images after build
+  command: docker images
+  register: build_images
+
+- name: Fail if httpd image not present
```
mike-nguyen: s/httpd/{{ db_image_name }}?

roles/docker_run/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- name: Run container with options
+  when: dr_run_options is defined
+  command: "docker run {{ dr_run_options }} --name {{ dr_image_name }} {{ dr_image_name }}"
+
+- name: Get running containers
+  command: "docker ps"
+  register: ps_run
+
+- name: Check for httpd container
```
mike-nguyen: s/httpd/{{ dr_image_name}}

roles/docker_rmi/docker_rmi_httpd_image/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Get the docker images against
+  command: docker images
+  register: images_after
+
+- name: Fail if the image is present
+  when: drmi_image_name in images_after.stdout
+  fail:
+    msg: |
+      Expected: {{ drmi_image_name }} is removed from docker images
+      Actual: docker images: {{ images_after.stdout }}
```
mike-nguyen: Is this file/directory misplaced?
`roles/docker_rmi/docker_rmi_httpd_image/tasks/main.yml` should just be `roles/docker_rmi/tasks/main.yml`

Same deal with the meta dir and main.yml file for docker_rmi.

mike-nguyen: Or were these supposed to be deleted?

miabbott: Crap.  Probably got things messed up when I rebased.  I'll sort it out.

tests/atomic/main.yml
```diff
@@ -415,7 +415,7 @@
     # Remove the locally tagged image so we can pull it again from the private
     # registry
     - role: docker_rmi
-      drmi_image: "{{ local_name }}:latest"
+      drmi_image_name: "{{ local_name }}:latest"
```
mike-nguyen: It looks like the docker_rmi role does not like the image name with tag being passed in because it looks for the whole string in the docker images output and the tag is separated in the docker images output.

```

TASK [docker_rmi : Fail if the image is not present] ***************************
    "changed": false, 
    "failed": true
}

MSG:
---
Expected: 172.17.0.1:5000/alpine:latest container is in docker images
Actual: docker images REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
docker.io/alpine         latest              dca2bcc01afd        1 seconds ago       4.143 MB
docker.io/busybox        latest              5b0d59026729        2 weeks ago         1.146 MB
docker.io/registry       2                   d1fd7d86a825        4 weeks ago         33.26 MB
172.17.0.1:5000/alpine   latest              3fd9065eaf02        4 weeks ago         4.143 MB
```

miabbott: Totally my fault, I didn't actually test these changes on the `atomic` test.  :\

I'll get it fixed tomorrow, likely.

mike-nguyen: The robots do the testing.  I just looked at the CI failure :laughing:

miabbott:
Pushed a new commit that I think will handle the tagged and non-tagged use cases.

miabbott:
CI checks are kind of useless here...

- fedora images are failing because the `docker` tests don't run on them
- `centos/7/atomic` is failing due to selinux denial

mike-nguyen:
The code looks good. Just to be sure I tested atomic/docker/improved-sanity-test against RHEL and it all passed!

PR #332, opened 2018-02-08
selinux_verify: add tags for each label check

miabbott:
The `selinux_verify` role performs multiple checks of files and
processes on the host to make sure all the SELinux labels are
correct.  However, if we wanted to exclude a particular part of the
role (because of known issues), we would have to exclude the entire
role.

This change introduces a set of tags assigned to different tasks in
the role, so that we are able to selectively choose which checks
should run as part of the role.

### DIFF
--- a/roles/selinux_verify/tasks/main.yml
+++ b/roles/selinux_verify/tasks/main.yml
@@ -1,6 +1,24 @@
 ---
 # vim: set ft=ansible:
 #
+# This role can be used to verify that various SELinux file labels and process
+# labels are correct on a system under test.  The role will check a set of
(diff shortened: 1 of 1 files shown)


miabbott:
@ashcrow Could you give this a quick review?  I want to be able to use these tags in our internal jobs sooner rather than later

ashcrow:
Looking

ashcrow:
LGTM. 

I suggest adding information about these tags and when to use them if they don't already exist in the documentation.

miabbott:
@ashcrow thanks for the review + documentation suggestion.  I've pushed a fixup with added docs.

PR #333, opened 2018-02-09
i-s-t: add test for non-priv systemd container

miabbott:
Since it is possible to run systemd in a container without the
`--privileged` flag, we decided we should have a test for this going
forward.

This change adds in the files necessary to build an `httpd` container
that uses `systemd` as its init mechanism.  Additionally, we add the
building/running of said container to the `improved-sanity-test`.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -366,6 +366,16 @@
       tags:
         - docker_rm_container_again
 
+    # Configure container_manage_cgroup boolean
+    # See - https://bugzilla.redhat.com/show_bug.cgi?id=1510139
+    #     - https://bugzilla.redhat.com/show_bug.cgi?id=1554881
+    - when: ansible_distribution == "RedHat" or
+            ansible_distribution == "Fedora"
+      role: selinux_boolean
+      g_seboolean: "container_manage_cgroup"
+      tags:
+        - container_manage_cgroup_bool
+
     # TEST
     # Build non-priviledged systemd httpd container
     - role: docker_build
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -375,6 +375,7 @@
       g_seboolean: "container_manage_cgroup"
       tags:
         - container_manage_cgroup_bool
+        - non_priv_systemd_httpd_container
 
     # TEST
     # Build non-priviledged systemd httpd container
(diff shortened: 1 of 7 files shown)


miabbott:
This requires #321, #330, and #331 to be merged before this can properly be reviewed.

Once those PRs are merged, I'll rebase this PR and formal review can begin.

miabbott:
One thing to note about this PR is that with [RHBZ#1510139](https://bugzilla.redhat.com/show_bug.cgi?id=1510139) still a problem, adding this test to `improved-sanity-test` is going to cause a lot of failures in our F27 CI checks (and in our internal F27 jobs).

We could put a `when:` statement around that part of the test until the BZ is fixed as a hacky workaround.

@dustymabe @mike-nguyen Any thoughts?

miabbott:
After waiting a long while for progress on the above mentioned BZ, it seems the issue is deeper than originally thought and is going to require a fix in the kernel.

The workaround for the time being is to enable the `container_manage_cgroup` SELinux boolean.

The last commit I just posted does this ahead of building/running the `httpd` systemd container and it seems to work across the board.  

Going to lift the WIP label from this as a result.

tests/improved-sanity-test/main.yml
```diff
@@ -366,6 +366,47 @@
...
+      tags:
+        - check_open_port_systemd
+
+    # TEST
+    # Remove the non-privileged systemd httpd container
+    - role: docker_rm_container
+      drc_container_name: "{{ g_osname }}_systemd_httpd"
+      tags:
+        - docker_rm_container_systemd
+
```
mike-nguyen: I just had a thought while reviewing the code.  Since this is actually a group of roles that test the non-priv systemd container, can we add an additional tag on each of them to group them together (for example adding a tag of `non-priv-systemd-test` to each of these roles)? If there is a bug causing a failure, we can just say `skip-tags non-priv-systemd-test`.  I think this can be done in other parts of the improved sanity test also.  What do you think?

miabbott: Extra tags is easy!  

(I actually had a similar thought and considered making a 'meta' role for this specific test case that just called out to individual roles.  I kind of like having each role available in the test explicitly, though, as it is that much clearer to understand what is happening.  We already have roles with true `meta` requirements, which is just additional layers of the onion to peel back when debugging)

mike-nguyen: I also like having each role available at the test level so its easier to figure out what's happening when there are failures.  Having tags is like extra metadata to see those roles are grouped.  Nice work on this PR!

PR #334, opened 2018-02-09
selinux_verify: update 'vars' files for RHEL

miabbott:
We had some older version of SELinux label information sitting around
in this repo, but I don't think much has changed in the RHELAH 7.4
time-frame.  Additionally, RHELAH 7.5 is coming soon and it was picking
up the stale SELinux label info.

If you are using this role against RHELAH 7.3 (or older), you're going
to have a bad time.


mike-nguyen:
LGTM.  I tried against 7.3 just for kicks and had a bad time. LOL.

PR #335, opened 2018-02-11
 k8s-cluster: fix test for F27AH

miabbott:
In Fedora 27 Atomic Host, all of the `kubernetes` related packages
were removed from the host.  The replacement functionality is now
provided by system containers.

This set of changes reworks the `kubernetes_setup` role to utilize the
new system containers to bring up a single node cluster and then
proceed with the remaininder of the `k8s-cluster` test.

NOTE:  the system containers are only used in the case where the
`kubernetes` packages are not present on the host.  For other
platforms where they are present on the host, the original method of
using manifests continues to be used.

Many thanks to Kedar Bidarkar for starting this work!

### DIFF
--- a/roles/k8_cluster_services_rc_setup/tasks/main.yml
+++ b/roles/k8_cluster_services_rc_setup/tasks/main.yml
@@ -3,80 +3,73 @@
 #  This role sets up the replication controller and services for the webserver
 #  and database server
 #
-  - name: Fail if ansible_docker0.ipv4.address is not defined
-    when: ansible_docker0.ipv4.address is not defined
-    fail:
-      msg: "ansible_docker0.ipv4.address it not defined!"
-    run_once: true
+- name: Fail if kubectl_path is not defined
+  when: kubectl_path is not defined
+  fail:
+    msg: |
+      This role requires that `kubectl_path` be defined before it can be used.
+
+- name: Fail if ansible_docker0.ipv4.address is not defined
+  when: ansible_docker0.ipv4.address is not defined
+  fail:
+    msg: "ansible_docker0.ipv4.address it not defined!"
+  run_once: true
 
 
-  - name: copy db-rc.yml, db-service.yml, webserver-rc.yml, webserver-service.yml
-    copy:
-      src=roles/k8_cluster_services_rc_setup/files/{{ item }}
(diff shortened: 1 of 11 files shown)


miabbott:
CI checks all green...

nailed it

roles/kubernetes_setup/tasks/main.yml
```diff
@@ -9,113 +9,173 @@
...
+      with_items:
+        - "kubelet"
+        - "kube-proxy"
+        - "kube-controller-manager"
+        - "kube-scheduler"
+
+- when: not use_system_containers
+  block:
+    - name: pull kubernetes api-server, controller-mgr, scheduler
+      when: not use_system_containers
```
mike-nguyen: Redudant?

roles/k8_cluster_services_rc_setup/tasks/main.yml
```diff
@@ -25,45 +25,57 @@
...
+          (ansible_distribution == "CentOSDev")
+    set_fact:
+      kubectl_path: "/usr/local/bin/kubectl"
+
+  - name: Define kubectl path for F26AH, Centos/RHEL
+    when: (ansible_distribution == "RedHat") or
+          (ansible_distribution == "Fedora" and ansible_distribution_major_version < '27') or
+          (ansible_distribution == "CentOS")
+    set_fact:
+      kubectl_path: "/usr/bin/kubectl"
```
mike-nguyen: Can we move these set_facts into a separate role and call them as a dependency of the role?  As the k8 rpms get removed from the streams, the logic will only need to be changed in a single role as opposed to several places.

mike-nguyen:
^^ This can be done in a separate PR.  I am okay with merging if you remove the redundant when clause in line 115.

miabbott:
The separate role for facts as a dependency is a good idea.  I'll work that up.

roles/k8_cluster_services_rc_setup/tasks/main.yml
miabbott: Ah dang it, I missed this piece in my latest commits....

mike-nguyen:
Nailed it again!  :100:  Great work as always!

PR #337, opened 2018-02-15
k8-cluster: add retries to final check

miabbott:
I've been seeing some failures in our internal jobs that look like the
final `curl` is timing out.  (Weird, since the connection is to
localhost...)

This adds some retries to the final check using `curl` to try to
alleviate the timeouts.


mike-nguyen:
:+1:

PR #338, opened 2018-02-15
tree: use F27 containers where possible

miabbott:
It would probably best to start using Fedora 27 containers instead of
Fedora 26.  While Fedora 26 will be 'maintained' until a month after
Fedora 28, I think we will get better support from the Fedora
community if we use the most recent release.

Additionally, I had to tweak the `atomic_images_delete_all` role to
remove any containers first before removing the images.  (This might
have been an artifact that came from re-using some test VMs, but
should be safe to do either way.)


cgwalters:
LGTM

PR #339, opened 2018-02-15
atomic_pull: add retry logic for pull

miabbott:
Noticed that this role was missing some retry logic and since
networking can be dicey, we should have some retry logic.


mike-nguyen:
LGTM

PR #340, opened 2018-02-16
roles: new role to validate firewalld install/config

### LINKED ISSUE #315: add test(s) for 'firewalld' on supported platforms
`firewalld` is starting to show up in various Fedora streams (FAHC, Rawhide, F27 Testing), so we should have a simple test that verifies it is operating correctly.

To start, we can just make sure it is enabled via `systemctl is-enabled firewalld`.  Then try adding/removing rules afterwards?

miabbott:
This introduces a new role which can be used to validate the
installation/configuration of the `firewalld` service.  It will
determine if the host under test should have `firewalld` installed and
how it should be configured if it is installed.

This initial attempt is very simple, but could be expanded to do more
complex checks if desired.

The role is used in the `improved-sanity-test` before the host is
upgraded and after upgrade.

Closes #315

### DIFF
--- a/roles/firewalld_service_verify/tasks/main.yml
+++ b/roles/firewalld_service_verify/tasks/main.yml
@@ -1,0 +1,57 @@
+---
+# vim: set ft=ansible:
+#
+# This role confirms that the 'firewalld' package/service is correctly
+# installed (or not installed) on the host under test.
+#
+- name: Default to firewalld being uninstalled
+  set_fact:
+    firewalld_installed: false
+
+- name: Determine if firewalld should be installed
+  when: (ansible_distribution == 'Fedora' and
+         ansible_distribution_major_version | version_compare('26', '>')) or
+        (ansible_distribution == 'RedHat' and
+         ansible_distribution_version | version_compare('7.4', '>'))
+  set_fact:
+    firewalld_installed: true
+
+- when: firewalld_installed
+  block:
+    - name: Verify firewalld package is installed
+      command: rpm -q firewalld
+
+    - name: Verify firewalld is disabled by default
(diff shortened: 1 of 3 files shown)


roles/firewalld_service_verify/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+
+- when: firewalld_installed
+  block:
+    - name: Verify firewalld package is installed
+      command: rpm -q firewalld
+
+    - name: Verify firewalld is disabled by default
+      command: systemctl is-enabled -q firewalld
+      register: is_enabled
+      failed_when: is_enabled.rc == 0
```
mike-nguyen: Discussion point.  Do you think its clearer to look at return codes or use jinja2 filtering (`is_enabled|success`).  The use seems mixed in this role and I'm just wondering if we should choose one or the other for consistency and clarity if it provides the same logic.

miabbott: Good point.  I used `enable_start|success` because the `service` task doesn't have an `rc` attribute.  I think I prefer the Jinja2 filter approach overall.

I'll make changes to standardize on that.

miabbott:
bot, retest this please

miabbott:
Nuts, I was hoping retesting this with the changes made to our PAPR setup would cause it to skip F26 and use the new config for `centos/7/atomic`

Anyways, those two are going to fail, but unrelated to the PR.  @mike-nguyen I think this is ready to get merged...do you agree?

mike-nguyen:
Yes, I took another quick look over and it looks good.  I'll merge it.

PR #341, opened 2018-02-20
k8s_set_facts: setup facts when kube node RPM not installed

miabbott:
The logic for determining the use of system containers was missing a
case when the `kubernetes-node` RPM was not installed on the host, but
the host was also not a Fedora/CentOS platform.  For example, our
internal 'autobrew' streams.

This updates the logic to handle this case.


mike-nguyen:
LGTM

PR #343, opened 2018-02-20
ci: enable extra args to be passed to .test_director

miabbott:
The `centos/7/atomic` context has been consistently failing because
there are AVC denials from the kernel that are present in the version
of CentOS AH we are using.  This is preventing any useful CI checks
from being run on that platform.

This change introduces a simple check to the `.test_director` script to
see if any arguments were passed to the script.  This way, we are able
to skip tags we know are causing problems in our CI checks.

In order to utilize this change, I re-arranged the order of contexts
in the `.papr.yml` file in order to maintain inheritence between the
contexts and only pass in the extra args to the `centos/7/atomic`
context.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -44,5 +44,5 @@ cluster:
 context: centos/7/atomic
 
 tests:
-  - ./test_director --skip-tags kernel_avc_denied
+  - ./.test_director --skip-tags kernel_avc_denied
 
--- a/.papr.yml
+++ b/.papr.yml
@@ -25,11 +25,11 @@ inherit: true
 cluster:
   hosts:
     - name: testnode
-      distro: centos/7/atomic
+      distro: centos/7/atomic/continuous
   container:
       image: registry.fedoraproject.org/fedora:27
 
-context: centos/7/atomic
+context: centos/7/atomic/continuous
 
 ---
 inherit: true
@@ -37,8 +37,12 @@ inherit: true
 cluster:
   hosts:
     - name: testnode
-      distro: centos/7/atomic/continuous
+      distro: centos/7/atomic
   container:
       image: registry.fedoraproject.org/fedora:27
 
-context: centos/7/atomic/continuous
+context: centos/7/atomic
+
+tests:
+  - ./test_director --skip-tags kernel_avc_denied
+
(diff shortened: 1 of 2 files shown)


miabbott:
@jlebon Could you have a look to make sure my assumptions about how PAPR parses the YAML file are correct?

.test_director
```diff
@@ -18,15 +23,15 @@ NEW_TESTS=( $(git diff --name-only $(git merge-base origin/master HEAD) | grep '
 # no new tests found, just run improved sanity test
 if [ ${#NEW_TESTS[@]} -eq 0 ]; then
     ansible-playbook --syntax-check tests/improved-sanity-test/main.yml
-    ansible-playbook -vi testnode, tests/improved-sanity-test/main.yml
+    ansible-playbook -vi testnode, tests/improved-sanity-test/main.yml $EXTRA_ARGS
```
jlebon: Minor: you can just do `"$@"` to directly pass through arguments. So then you wouldn't have to quote the args in the YAML, e.g.:

```yaml
tests:
  - ./test_directory --skip-tags kernel_avc_denied
```

jlebon:
Your assumptions are correct. :)

.papr.yml
```diff
@@ -37,20 +37,24 @@ inherit: true
...
-      distro: centos/7/atomic/continuous
+      distro: centos/7/atomic
   container:
       image: registry.fedoraproject.org/fedora:26
 
-context: centos/7/atomic/continuous
+context: centos/7/atomic
+
+tests:
+  - ./test_director '--skip-tags kernel_avc_denied'
```
jlebon: This is missing a `.`, no?

.test_director
miabbott: Oh word, so I don't even have to check for the existence of the args.

.papr.yml
miabbott: Good catch...I was just flying through this in a fit of rage  :laughing:

miabbott:
Well, had to force push my rebase  ¯\_(ツ)_/¯

mike-nguyen:
This is a great idea!

jlebon:
LGTM!

PR #348, opened 2018-02-22
use 'rpm_ostree_status' role for all the JSON output

miabbott:
We had a bunch of instances in the tree where we were doing
`rpm-ostree status --json` and then going through the motions to parse
the JSON to determine things like the booted deployment, etc.
Additionally, we were littering our test logs with the 'noisy' keys
from the JSON output that were handled in #301.

This change rips out all the invocations of `rpm-ostree status --json`
and replaces them with the usage of the `rpm_ostree_status` role.  Now
we have a consistent way of dealing with the JSON output from the
`status` command and only one place where we have to make changes if
needed.


mike-nguyen:
This looks good to me.  Nice work on cleaning up the when statements also.  I need to be more conscious of looking at this while I make PRs in the roles

PR #350, opened 2018-03-02
roles: adapt to Rawhide == F29

miabbott:
We have to bump our $releasever for Rawhide now that F28 has branched,
which means Rawhide == F29.

I've added the necessary files to cover this change as well as
rearranged some things under `selinux_verify/vars` to make a little
more sense (I hope).

While I was mucking around in here, I continued to remove references
to older versions of Fedora (24,25) where possible.


mike-nguyen:
The code looks good to me but there's a failure in fedora/27/atomc with AVC denials.  It seems unrelated to the changes.

miabbott:
> The code looks good to me but there's a failure in fedora/27/atomc with AVC denials. It seems unrelated to the changes.

Strange...our sanity tests should have caught that before now.  I'm going to retry that test to see what happens again.

miabbott:
LOL...fixed it.  :laughing:

mike-nguyen:
:+1:  Merging!

PR #351, opened 2018-03-06
README: add CentOS AH results

miabbott:
We've enabled the testing of CentOS AH on our internal Jenkins
instance, so let's start showing some results here too.


miabbott:
The fully rendered version can be seen here - https://github.com/miabbott/atomic-host-tests/tree/add_centosah_dashboard

mike-nguyen:
LGTM!

PR #352, opened 2018-03-07
README: remove f26 results + spelling correction

miabbott:
Remove the F26 stable results and correct 'Centos'


miabbott:
I'm merging this without review because it is just the README and also because YOLO

PR #353, opened 2018-03-08
Add git to tools container

mike-nguyen:
For the rpm-ostree compose test, we need the ability to checkout
a git repo with the necessary files to compose a tree.  Instead of
package layering git onto the host, we can use the tools container
with git installed to clone the repo and save it to a volume mount
shared between the host and container.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -2,6 +2,7 @@ Small tools container for use with [atomic-host-tests](https://github.com/projec
 
 As of 09-Jan-2017, the following utilities/packages are installed
   - `jq`
+  - `git`
 
 To run the container:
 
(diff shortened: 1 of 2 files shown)


tools/README.md
```diff
@@ -2,6 +2,7 @@ Small tools container for use with [atomic-host-tests](https://github.com/projec
 
 As of 09-Jan-2017, the following utilities/packages are installed
```
miabbott: nit: could you update the date here?

mike-nguyen: Done

tools/README.md
```diff
@@ -1,6 +1,6 @@
 Small tools container for use with [atomic-host-tests](https://github.com/projectatomic/atomic-host-tests)
 
-As of 09-Jan-2017, the following utilities/packages are installed
+As of 09-Mar-2018, the following utilities/packages are installed
```
miabbott: Just realized I had the wrong year in there to begin with.  :laughing:

PR #354, opened 2018-03-12
k8-cluster:  revert to using older system containers

miabbott:
Two major changes here:

1)  Pin the system containers to a previous version that was working

2)  Make sure that 'atomic 1.22' is installed on the host

The new requirement on 'atomic 1.22' is to resolve the following issue
with SELinux labeling - projectatomic/atomic#1185

This is only a temporary change; once the latest versions of the
system containers are known to be working, we can revert a lot of
this.

### DIFF
--- a/tests/k8-cluster/main.yml
+++ b/tests/k8-cluster/main.yml
@@ -21,6 +21,39 @@
   hosts: all
   become: true
 
+# This giant mess of pre-tasks is here to workaround the problem that is fixed
+# in this PR - https://github.com/projectatomic/atomic/pull/1185
+#
+# The fallout from that missing functionality is that the 'kube-proxy' system
+# container will not start correctly unless atomic 1.22 is used to install
+# the system container with the right SELinux labels.
+#
+  pre_tasks:
+    - include_role:
+        name: rpm_version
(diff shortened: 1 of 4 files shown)


tests/k8-cluster/main.yml
```diff
@@ -21,6 +21,39 @@
...
+    - include_role:
+        name: rpm_version
+      vars:
+        rv_rpms: atomic
+
+    - when:
+        - ansible_distribution == 'Fedora'
+        - ansible_distribution_major_version | version_compare('26', '>')
+        - g_atomic_host is defined
+        - g_atomic_host['atomic'] | version_compare('1.22', '<=')
```
mike-nguyen: Does CAHC has atomic >= 1.22 so it doesn't need this step here?

miabbott: The logical AND should limit this to just Fedora hosts that are after F26 with a version of atomic less than 1.22

mike-nguyen: :+1:

mike-nguyen:
LGTM.  Nice :male_detective: work on tracking down the root cause.

PR #356, opened 2018-03-12
k8-cluster: fix missing double quotes on regex

mike-nguyen:
I'm not sure how this test was working but there are two regex
groups that were missing double quotes based on how the statements
were structured.

The first regex was a replace which ends in .*$ which will match
anything including the end quote but the replacement string did
not include the end double quote.

The second regex was a regex group capture that does not include
the end double quote so the line replacement needs to have the
double quote included.


mike-nguyen:
Needs #354 merged first before CI checks will pass.

miabbott:
regex...how do they work?

Good catch!

PR #358, opened 2018-03-14
papr: skip default_t checks temporarily for CAHC

mike-nguyen:
Lets not fail the CAHC CI check because a package isn't in CAHC yet.
Since this label isn't fatal to functionality, it's pretty safe
to skip this until we get an updated version of 'container-selinux'
with the neccessary fix:

https://github.com/projectatomic/container-selinux/pull/50

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -31,6 +31,10 @@ cluster:
 
 context: centos/7/atomic/continuous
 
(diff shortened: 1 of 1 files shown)


miabbott:
Just made #361 to address this.  :)

PR #360, opened 2018-03-14
system-containers: use FQ container name

mike-nguyen:
When a fully qualified container name is not used, skopeo will use
the default container registry and use a short shorname.  In this
case gscrivano/hello-world shows up as hello world in the atomic
images list output.  The system-container looks for the short name
that was passed in when installing the system container so the test
ends up failing.  The simple fix is to use fully qualified container
names.


miabbott:
LGTM!

I wonder if there are other places in the repo where we need to use fully qualified names

PR #361, opened 2018-03-14
papr: skip the 'default_t' selinux check on CAHC

miabbott:
We are waiting for the following change to land in `container-selinux`
to alleviate some of our checks from failing:

https://github.com/projectatomic/container-selinux/pull/50

Because the error is not fatal, let's skip this check for now.


miabbott:
Closed in favor of #358

mike-nguyen:
haha samesies

PR #363, opened 2018-03-16
README: add notes about package requirements

### LINKED ISSUE #267: give better example of installing requirements on F27
I tried to use the `improved-sanity-test` from within a F26 Docker container and found that I had to perform a number of RPM installs before the `pip install -r requirements.txt` command would complete successfully.

While not a strict set of requirements for using this repo, it would probably useful to mention this kind of additional setup in the `README` or in the other docs we have.

As a reference, here's what my `dnf` command history looked like:

```
    1  dnf -y install git python2-virtualenv
   12  dnf install gcc
   14  dnf install /usr/lib/rpm/redhat/redhat-hardened-cc1
   20  dnf -y install sshpass
   22  dnf -y install rsync
```

miabbott:
This adds a short note about packages that need to be installed ahead
of time before you can use the `virtualenv` approach for setting up
an a-h-t environment.

Closes #267

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,19 @@ interfere with your current workspace.
 To setup a virtualenv, follow the steps below after cloning atomic-host-tests:
 
 ```
-pip install virtualenv
 virtualenv my_env
 source my_env/bin/activate
 pip install -r requirements.txt
 ```
+
+The above instructions assume you have some existing tooling installed on your
+workstation.  For example, on a Fedora 27 system the following packages needed
+to be installed ahead of time:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,12 +146,20 @@ interfere with your current workspace.
...
 
 ```
-pip install virtualenv
 virtualenv my_env
 source my_env/bin/activate
 pip install -r requirements.txt
 ```
 
+The above instructions assume you have some existing tooling installed on your
+workstation.  For example, on a Fedora 27 system the following packages needed
```
mike-nguyen: s/needed/need

mike-nguyen:
LGTM

PR #364, opened 2018-03-16
rpm_sig_verify: update GPG keys for F28/Rawhide

miabbott:
Now that F28 has branched, we need to use the proper GPG keys that are
used for signing the RPMs


mike-nguyen:
LGTM!

PR #365, opened 2018-03-16
docker_storage_verify: add in support for rhelah 7.5

miabbott:
This modifies the `docker_storage_verify` role to look for the
`ansible_distribution_version` in the name of the var file.  The
RHELAH 7.5 version of said file is also added.


mike-nguyen:
LGTM!

PR #366, opened 2018-03-22
Install rpms for rpm-ostree install gpg test

mike-nguyen:
Install createrepo_c and python on the aht-tools container so it we
can create a local repo and serve it through python's
SimpleHTTPServer.

Also added notes in the README to explain why the packages were added.
This might help in the future to determine if a package is still
needed in the tools container.


miabbott:
The CI checks won't actually pick up the new version of the container until *after* the commit is made, so I'm just going to LGTM this and watch what happens.

mike-nguyen:
:+1:

Issue #349, opened 2018-02-23
create a test to verify GPG checking on 'rpm-ostree install'


miabbott:
We have a manual test case internal to Red Hat that grabs an unsigned RPM, props up a yum repo with it, and validates that GPG checking is enforced.

We can do something similar with the `aht-dummy` RPM that exists in the repo.

1.  Grab RPM
2.  `createrepo_c` on a directory
3.  create `/etc/yum.repos.d/dummy.repo` with `gpgcheck=1`
4.  try to `rpm-ostree install` and confirm it fails
5.  change `gpgcheck=0`
6.  try to `rpm-ostree install` and confirm it succeeds

mike-nguyen:
#367

miabbott:
Closed via #367

PR #367, opened 2018-03-22
Add gpg package layering test

mike-nguyen:
This commit adds a gpg test for package layering.  The test uses the
aht-tools container to create a local repo with the aht-dummy rpm and
serve it through SimpleHTTPServer.  The tools container was chosen
because it was an easy way to host the repo without requiring another
system or installing packages on the host.

Addresses issue #349

### DIFF
--- a/tests/pkg-layering/main.yml
+++ b/tests/pkg-layering/main.yml
@@ -339,7 +339,7 @@
 
     - name: Add file contents to local.repo with gpgcheck enabled
       blockinfile:
-         path: /etc/yum.repos.d/local.repo
+         dest: /etc/yum.repos.d/local.repo
          marker: ''
          block: |
            [local]
@@ -368,7 +368,7 @@
 
     - name: Modify local.repo to disable gpg checking
       replace:
-         path: /etc/yum.repos.d/local.repo
+         dest: /etc/yum.repos.d/local.repo
          regexp: 'gpgcheck=1'
          replace: 'gpgcheck=0'
 
--- a/tests/pkg-layering/main.yml
+++ b/tests/pkg-layering/main.yml
@@ -307,6 +307,93 @@
       vars:
         package: "{{ g_epel_pkg }}"
 
+- name: Package Layering - GPG Test
+  hosts: all
+  become: true
+
+  tags:
+    - gpg
+
+  tasks:
+    - name: Make /srv/localrepo directory
+      file:
+        path: /srv/localrepo
+        state: directory
+        mode: 0755
+
+    - name: Copy aht-dummy rpm to /srv/localrepo
+      copy:
+         src: ../../rpm/aht-dummy-1.0-1.noarch.rpm
+         dest: /srv/localrepo
+         owner: root
+         group: root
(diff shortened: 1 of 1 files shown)


mike-nguyen:
Dependent on #366

tests/pkg-layering/main.yml
```diff
@@ -307,6 +307,93 @@
...
+  tasks:
+    - name: Make /srv/localrepo directory
+      file:
+        path: /srv/localrepo
+        state: directory
+        mode: 0755
+
+    - name: Copy aht-dummy rpm to /srv/localrepo
+      copy:
+         src: ../../rpm/aht-dummy-1.0-1.noarch.rpm
```
miabbott: Could you use `playbook_dir` here instead of the relative path?

http://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#magic-variables-and-how-to-access-information-about-other-hosts

tests/pkg-layering/main.yml
```diff
@@ -307,6 +307,93 @@
...
+  tasks:
+    - name: Make /srv/localrepo directory
+      file:
+        path: /srv/localrepo
+        state: directory
+        mode: 0755
+
+    - name: Copy aht-dummy rpm to /srv/localrepo
+      copy:
+         src: '{{ playbook_dir}}/../../rpm/aht-dummy-1.0-1.noarch.rpm'
```
miabbott: LOL...that is just as ugly as before.  But I guess it is a bit more deterministic.

tests/pkg-layering/main.yml
```diff
@@ -307,6 +307,93 @@
...
+    - name: Wait for port 80 to open
+      wait_for:
+        port: 80
+        delay: 5
+        timeout: 30
+
+    - name: Attempt to install unsigned rpm with gpgcheck enabled
+      command: rpm-ostree install aht-dummy
+      register: ros_install
+      failed_when: ros_install|success
```
miabbott: Should there be an additional check here to see if the error message returned is about GPG checking?

tests/pkg-layering/main.yml
mike-nguyen: Haha, yeah we still have to go back up two directories to get to the rpm directory.  I could symlink into the directory to make it cleaner.  WDYT?

tests/pkg-layering/main.yml
mike-nguyen: I can add that.

mike-nguyen:
Pushed a fixup with a check for the error message

tests/pkg-layering/main.yml
miabbott: Let's leave it as is for now.  It would be nice if we could figure out a way to derive the root of the repo while running tests and set a fact.

miabbott:
> Pushed a fixup with a check for the error message

I think you forgot the push...the last commit was from 4 days ago

mike-nguyen:
entered the wrong pw lol

miabbott:
LGTM!

PR #368, opened 2018-03-23
tree: initial changes to support Fedora STI/STR

miabbott:
These two commits include the initial changes to try to support the [Fedora Standard Test Interface](https://fedoraproject.org/wiki/CI/Tests) and [Standard Test Roles](https://fedoraproject.org/wiki/CI/Standard_Test_Roles).

The first commit makes small changes to the `improved-sanity-test` so that it can be invoked by the Fedora STI.

The second commit changes the methodology of the `reboot` role to handle the way the STI boots up an Atomic Host as a test subject.  

There are more details in the individual commit messages and I encourage you to read them.

(I'm not 100% confident about getting these merged in, so marking it as WIP for now)

### DIFF
--- a/roles/reboot/tasks/main.yml
+++ b/roles/reboot/tasks/main.yml
@@ -17,7 +17,6 @@
 - set_fact:
     real_ansible_host: "{{ ansible_host }}"
     timeout: "{{ cli_reboot_timeout | default('120') }}"
-    wf_delay: "{{ cli_wf_delay | default('30') }}"
 
 # Have to account for both because Fedora STR uses the old version of these
 # inventory values for some reason.
@@ -49,7 +48,7 @@
     wait_for host={{ real_ansible_host }}
     port={{ real_ansible_port | default('22') }}
     state=started
-    delay={{ wf_delay }}
+    delay=30
     timeout={{ timeout }}
     search_regex="OpenSSH"
   become: false
--- a/roles/reboot/tasks/main.yml
+++ b/roles/reboot/tasks/main.yml
@@ -17,6 +17,21 @@
 - set_fact:
     real_ansible_host: "{{ ansible_host }}"
     timeout: "{{ cli_reboot_timeout | default('120') }}"
+    wf_delay: "{{ cli_wf_delay | default('30') }}"
+
+# Have to account for both because Fedora STR uses the old version of these
+# inventory values for some reason.
+- when: ansible_port is defined
+  set_fact:
+    real_ansible_port: "{{ ansible_port }}"
+
+- when: ansible_ssh_port is defined
+  set_fact:
+    real_ansible_port: "{{ ansible_ssh_port }}"
+
+- name: Get original bootid
+  command: cat /proc/sys/kernel/random/boot_id
+  register: orig_bootid
 
 - name: restart hosts
   when: (not skip_shutdown is defined) or (not skip_shutdown)
@@ -29,17 +44,19 @@
 # that they don't use sudo, which may require a password, and is not necessary
 # anyway.
 
-- name: wait for hosts to go down
-  local_action:
-    wait_for host={{ real_ansible_host }}
-    port=22 state=absent delay=1 timeout={{ timeout }}
-  become: false
-
 - name: wait for hosts to come back up
   local_action:
     wait_for host={{ real_ansible_host }}
-    port=22 state=started delay=30 timeout={{ timeout }}
+    port={{ real_ansible_port | default('22') }} state=started delay={{ wf_delay }} timeout={{ timeout }}
   become: false
+
+# I'm not sure the retries are even necessary, but I'm keeping them in
+- name: Wait until bootid changes
+  command: cat /proc/sys/kernel/random/boot_id
+  register: new_bootid
+  until: new_bootid.stdout != orig_bootid.stdout
+  retries: 6
+  delay: 10
 
 # provide an empty iterator when a list is not provided
 # http://docs.ansible.com/ansible/playbooks_conditionals.html#loops-and-conditionals
--- a/roles/reboot/tasks/main.yml
+++ b/roles/reboot/tasks/main.yml
@@ -47,7 +47,11 @@
 - name: wait for hosts to come back up
   local_action:
     wait_for host={{ real_ansible_host }}
-    port={{ real_ansible_port | default('22') }} state=started delay={{ wf_delay }} timeout={{ timeout }}
+    port={{ real_ansible_port | default('22') }}
+    state=started
+    delay={{ wf_delay }}
+    timeout={{ timeout }}
+    search_regex="OpenSSH"
   become: false
 
 # I'm not sure the retries are even necessary, but I'm keeping them in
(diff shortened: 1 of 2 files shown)


miabbott:
If you want to test this, this was my approach:

1.  On your workstation, install `oci-kvm-hook`.  This will allow you to boot VMs from within a container.
2.  Run a Fedora 27 container
3.  In the container, install the following:

`# dnf -y install ansible git python2-dnf libselinux-python qemu rsync standard-test-roles`

4.  Grab a copy of the latest Fedora AH qcow2 - https://getfedora.org/atomic_qcow2_latest
5.  Setup env vars:

```
# export ANSIBLE_INVENTORY=$(test -e inventory && echo inventory || echo /usr/share/ansible/inventory)
# export TEST_SUBJECTS=/path/to/Fedora-Atomic-27-20180314.0.x86_64.qcow2
```

6.  Checkout the PR
7. Run the playbook:

`# ansible-playbook -v --tags atomic tests/improved-sanity-test/main.yml -e cli_wf_delay=60`

mike-nguyen:
Nice work so far.  It worked for me

mgahagan73:
It is working for me, I see boot/console logs for the atomic host under test under artifacts/ so I think the last remaining piece is to get the rest of the test logs under artifacts.

mgahagan73:
I think we can add a tests.yml like the one below as well.

root@amd-pike-01 atomic-host-tests]# less tests.yml 
- include: tests/improved-sanity-test/main.yml
  environment:
    TEST_ARTIFACTS: ./artifacts
    AHT_RESULT_FILE: "{{ TEST_ARTIFACTS }}/improved-sanity-test.log"
    cli_wf_delay: 60
  tags:
    atomic

miabbott:
@mgahagan73 I've been experimenting with that `tests.yml` file, but it doesn't seem like the `environment` settings are getting passed up to the 'include`.

I'm thinking I'll have to make further changes to `improved-sanity-test/main.yml` itself to use those variables.

mgahagan73:
I did find through more testing I had to declare cli_wf_delay as a var and not an environment variable to avoid failures due to the VM not being reachable. After my test runs I'm seeing the artifacts/ gets created with the boot log for the atomic host and the failure log being generated. My failure log gets created as an empty file when everything passes.  This is all with the ansible 2.4 rpm shipped with F27 so maybe we are seeing differences between versions. I know the README mentions 2.2 is the version the atomic host playbooks were developed for.

[root@amd-pike-01 ~]# ansible --version
ansible 2.4.3.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.14 (default, Mar 14 2018, 13:36:31) [GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
[root@amd-pike-01 ~]# rpm -q ansible
ansible-2.4.3.0-1.fc27.noarch

miabbott:
@mgahagan73 I'm running the playbook from inside a container and it has `ansible-2.4.3.0-1.fc27.noarch` installed.

I was seeing the log from the VM in the artifacts directory, but wasn't seeing any test log.

I might just merge these changes as is and then iterate towards the `tests/tests.yml` paradigm in a future PR.

mgahagan73:
I'm only seeing a test log when I get a failure. Setting cli_wf_delay to a low value seems to be a pretty easy way to induce a failure which gave me a test log based on whatever AHT_RESULT_FILE is set to. I can't seem to redefine the artifacts variable from either in the playbook or on the ansible command line which I suspect to be an issue with standard-test-roles.

roles/reboot/tasks/main.yml
```diff
@@ -29,18 +44,20 @@
...
-  become: false
-
 - name: wait for hosts to come back up
   local_action:
     wait_for host={{ real_ansible_host }}
-    port=22 state=started delay=30 timeout={{ timeout }}
+    port={{ real_ansible_port | default('22') }} state=started delay={{ wf_delay }} timeout={{ timeout }}
   become: false
 
+# I'm not sure the retries are even necessary, but I'm keeping them in
```
cgwalters: I started coding this too a bit before but I think the problem is we actually need to retry *both* the ssh connection and the boot_id check.  The way to think of this is: it was still racy before the STR approach since we were relying on the shutdown actually finding the ssh port closed, but we could (if the guest rebooted fast enough) have it always appear open. So we need to loop over both tasks.  This led me to https://stackoverflow.com/questions/30785281/one-loop-over-multiple-ansible-tasks

mike-nguyen:
> I'm only seeing a test log when I get a failure. Setting cli_wf_delay to a low value seems to be a pretty easy way to induce a failure which gave me a test log based on whatever AHT_RESULT_FILE is set to. I can't seem to redefine the artifacts variable from either in the playbook or on the ansible command line which I suspect to be an issue with standard-test-roles.

The AHT_RESULT_FILE  only captures the task/role that failed.  It does not contain contain the output of the test run.

miabbott:
> The AHT_RESULT_FILE  only captures the task/role that failed.  It does not contain contain the output of the test run.

Doh!  Embarrassing when you forget how your own tests work...

So there is going to be some additional effort required to capture the entire log.  Not sure if we want to do it natively as part of the callback plugin or just wrap the invocation of `ansible_playbook` in a shell script to capture the stdout.  Any thoughts here?

Like I said before, I think I'd prefer to get this PR merged and iterate in other PRs towards a more full-fledged adoption of the STI.

roles/reboot/tasks/main.yml
miabbott: @cgwalters Did you have to do this in any of the ostree tests you've been working on?

What's the idea for the implementation?  Break out those tasks into a separate file and loop over the `include` of the file?

miabbott: @mike-nguyen found this post which discusses the problem of rebooting a Vagrant box and the SSH port is forwarded to a non-standard port (much like the STI does):

https://blog.jeffli.me/blog/2017/05/19/a-pitfall-of-vagrant-when-using-ansibles-wait-for-module/

TL;DR - use `search_regex` with the `wait_for` module

I'm going to try to play with this approach sometime today.

miabbott:
The last commit I made appears to have improved the reliability of the `reboot` role, so I'm lifting the WIP label for this set of changes.  As stated previously, I'll work on additional changes required to support the STI.

roles/reboot/tasks/main.yml
```diff
@@ -17,6 +17,21 @@
 - set_fact:
     real_ansible_host: "{{ ansible_host }}"
     timeout: "{{ cli_reboot_timeout | default('120') }}"
+    wf_delay: "{{ cli_wf_delay | default('30') }}"
```
mike-nguyen: I just tested this with the no cli_wf_delay passed in and it worked nicely!  Is there still a need to be able to pass in a wf_delay with the regex_search?

miabbott: Aside: I actually tested with `cli_wf_delay=10` and it still worked well, so the regex is proving useful.

I'm not opposed to dropping out the option to configure the delay.

mike-nguyen:
LGTM, As planned I will merge this and then subsequent STI work can be in another PR.

Issue #336, opened 2018-02-13
migrate everything to use Ansible 2.4


miabbott:
In #287, I did some investigation into the work required to use Python3 with our tests.

Thankfully, most of the roles/tests did not have any huge issues.  However, the easiest way forward is to move to a newer version of Ansible, namely 2.4

The biggest reason to move to Ansible 2.4 is support.  Ansible 2.2 will be unsupported sooner rather than later (https://docs.ansible.com/ansible/2.4/release_and_maintenance.html).  While it is unlikely we are going to encounter a situation where we need to fix something in Ansible itself, I'd rather avoid that scenario altogether.

In order to get the repo fully supported by Ansible 2.4, we need the following:
  - switch to Ansible 2.4.x (Ansible 2.4.0.0 for now, due to the above comment)
  - stop using `include:` and start using `include_tasks:`

miabbott:
I think this might have a higher priority since the [Fedora Standard Test Roles/Interface]9https://fedoraproject.org/wiki/CI/Tests) also use Ansible 2.4.

Going to tag it into our Jira board.

miabbott:
We did a lot of work in #372, #373, #386, #389, and maybe some other PRs to handle this.

PR #372, opened 2018-04-02
roles: remove use of 'include:' statements

miabbott:
This PR removes the use `include:` across the roles to use `include_tasks` or `include_role`

The most notable change is moving the `set_is_atomic` role from the `common` directory to the `roles` directory.

The `redhat_subscription` role looks like it changed, but it really just got re-indented.

### DIFF
--- a/roles/ostree_admin_unlock/tasks/main.yml
+++ b/roles/ostree_admin_unlock/tasks/main.yml
@@ -7,7 +7,8 @@
 - name: Unlock system
   command: ostree admin unlock
 
-- include: '../../rpm_ostree_status/tasks/main.yml'
+- include_role:
+    name: rpm_ostree_status
 
 - name: Fail if unlocked is not set to development
   when: "'development' not in ros_booted['unlocked']"
--- a/roles/ostree_admin_unlock/tasks/main.yml
+++ b/roles/ostree_admin_unlock/tasks/main.yml
@@ -7,7 +7,7 @@
 - name: Unlock system
   command: ostree admin unlock
 
-- include_role:
+- import_role:
     name: rpm_ostree_status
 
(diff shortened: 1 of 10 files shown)


miabbott:
This partially addresses #336.  I'm going to work on updating the tests in a separate PR.

roles/ostree_admin_unlock/tasks/main.yml
```diff
@@ -7,7 +7,8 @@
 - name: Unlock system
   command: ostree admin unlock
 
-- include: '../../rpm_ostree_status/tasks/main.yml'
+- include_role:
```
mike-nguyen: Let's only use import_role as a baseline and only use include_role where we need it dynamically called.  I ran into a recursion depth limit using include.

roles/ostree_admin_unlock_hotfix/tasks/main.yml
```diff
@@ -6,7 +6,8 @@
 - name: Unlock system with hotfix
   command: ostree admin unlock --hotfix
 
-- include: '../../rpm_ostree_status/tasks/main.yml'
+- include_role:
```
mike-nguyen: Same here.

roles/rpm_version/tasks/main.yml
```diff
@@ -7,6 +7,6 @@
 # This role takes a list of rpms, retrieves the version, and stores it in a
 # dict named g_atomic_host
 #
-- include: version.yml
+- include_tasks: version.yml
```
mike-nguyen: import_tasks

mike-nguyen:
Probably need requirements.txt updated for this too

miabbott:
I'm just going to close this in favor of all the changes in #373

PR #374, opened 2018-04-04
rpm-ostree: add override tests

### LINKED ISSUE #345: expand 'rpm-ostree' test suite to include coverage for 'override' functionality
Since `override` has been out of experimental for a while, we should definitely have some coverage for it.

Make sure we cover:

`rpm-ostree override replace`
`rpm-ostree override remove`
`rpm-ostree override reset`

mike-nguyen:
The rpm-ostree override functionality has been out of experimental for
a while now so this commit adds tests for replace, remove, and reset.

These tests replace and remove packages that are common to RHEL,
Fedora, and CentOS Atomic Hosts.  The goal was to test against an
ostree commit release without modification.  The caveat to this approach
is that the package set may change and we will have to find another
common package to use.  If this becomes too troublesome, we can adopt
the strategy that the rpm-ostree unit tests uses and create a new
branch with a known rpm to replace and remove.

Closes #345


miabbott:
> centos/7/atomic is so old it doesn't have override functionality. I will put in a check for it

Weird, it should have parity with RHELAH.  Maybe it is still `rpm-ostree ex override`?

miabbott:
Ah, the `centos/7/atomic` image in the internal OpenStack instance was from Jan 9.

I've just updated it and will re-run the failed CI check again.

mike-nguyen:
I just downloaded and checked the latest Centos AH host image and it has the override.  It would have been quicker/easier to check the OpenStack image :-\

mike-nguyen:
bot, retest this please

mike-nguyen:
bot, retest this please

mike-nguyen:
bot, retest this please

miabbott:
@mike-nguyen Should the test be updated to use `import_*` instead of `include_*` where possible?

miabbott:
@mike-nguyen Could the multiple roles be combined into a single `override` role that would take one of the subcommands/verbs as an arg?

mike-nguyen:
@miabbott pushed changes to use one role with subcommands

miabbott:
> I was actually holding off on that until we actually moved to 2.4. This PR and the 2.4 have been in limbo for a bit. If the 2.4 stuff gets merged before this PR, I'll update this one. Otherwise I would rather make another PR after the 2.4 hits.

Makes sense.

Everything else looks good, so I'm going to merge this.

PR #379, opened 2018-04-19
papr: adjust which tests are skipped per platform

miabbott:
We want to run the 'docker' test suite in CI with
'g_docker_latest=false' for Fedora because the `docker-latest` package
is not shipped on that distro.

Also add the 'default_t' test to the list of skipped tests for CentOS
until the new version of atomic lands.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -17,7 +17,9 @@ packages:
   - rsync
 
 tests:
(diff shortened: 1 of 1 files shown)


mike-nguyen:
LGTM

PR #381, opened 2018-04-24
rpm-ostree/main: Run ostree admin status too

cgwalters:
Would have fallen over for https://bugzilla.redhat.com/show_bug.cgi?id=1571264


miabbott:
I was going to put this in the sanity test, but this works too.

PR #373, opened 2018-04-04
tree: changes needed to support Ansible 2.4

miabbott:
The main intention of these changes is to get the tests to a point where they can be used with Ansible 2.4 without any errors/warnings.  In order to get there, I had to update some of the roles along the way.

Specifically, I updated the `atomic` install/uninstall related roles to use unique variables names to prevent collisions during tests.  I also updated the `rpm_version` role to be more specific where tasks were being included from and how variables were being passed to `include_tasks`.

With those changes in place, I was able to update the tests to use `include_tasks` where a bare `include` was used.

PR #383, opened 2018-04-26
atomic/main: cleanup after the tests are done

miabbott:
The test was sloppy about cleaning up after itself, in that it did no
cleanup at all really.  This adds a final playbook that will cleanup
the containers/images and rollback any layering of 'atomic-registries'
package that might have occurred.

### DIFF
--- a/tests/atomic/main.yml
+++ b/tests/atomic/main.yml
@@ -468,3 +468,46 @@
     - role: command
       cmd: atomic images update {{ local_name }}:latest
       output: "Writing manifest to image destination"
+
+
+- name: Atomic- Cleanup
+  hosts: all
+  become: true
+
+  tags:
+    - cleanup
+
+  tasks:
+    - name: Check if 'atomic-registries' was layered in
+      shell: >
+        rpm-ostree status --json |
+        docker run -i docker.io/miabbott/aht-tools
+        jq '.deployments[0]["base-commit-meta"]["rpmostree.packages"]'
+      register: ros
+
+    - name: Indicate 'atomic-registries' is layered
+      when: "'atomic-registries' in ros.stdout"
+      set_fact:
+        ar_is_layered: true
+
+    - import_role:
+        name: docker_remove_all
+
+    # If the 'atomic-registries' package was layered, we can rollback,
+    # reboot, and cleanup to be back to pristine condition
+    - block:
+        - import_role:
+            name: rpm_ostree_rollback
+
+        - import_role:
+            name: reboot
+
+        - name: Cleanup deployments
+          command: rpm-ostree cleanup -rpmb
+      when:
+        - ar_is_layered is defined
+        - ar_is_layered
+
+    - when: ansible_distribution == 'RedHat'
+      import_role:
+        name: redhat_unsubscribe
--- a/tests/atomic/main.yml
+++ b/tests/atomic/main.yml
@@ -478,15 +478,13 @@
     - cleanup
 
   tasks:
-    - name: Check if 'atomic-registries' was layered in
-      shell: >
-        rpm-ostree status --json |
-        docker run -i docker.io/miabbott/aht-tools
-        jq '.deployments[0]["base-commit-meta"]["rpmostree.packages"]'
-      register: ros
+    # Run 'rpm-ostree status --json' to determine the current booted
+    # deployment and setup the 'ros_booted' variable
+    - import_role:
+        name: rpm_ostree_status
 
     - name: Indicate 'atomic-registries' is layered
-      when: "'atomic-registries' in ros.stdout"
+      when: "'atomic-registries' in ros_booted['base-commit-meta']['rpmostree.packages']"
       set_fact:
         ar_is_layered: true
 
@@ -495,7 +493,10 @@
 
     # If the 'atomic-registries' package was layered, we can rollback,
     # reboot, and cleanup to be back to pristine condition
-    - block:
+    - when:
+        - ar_is_layered is defined
+        - ar_is_layered
+      block:
         - import_role:
             name: rpm_ostree_rollback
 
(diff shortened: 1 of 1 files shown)


tests/atomic/main.yml
```diff
@@ -471,3 +471,46 @@
...
+      shell: >
+        rpm-ostree status --json |
+        docker run -i docker.io/miabbott/aht-tools
+        jq '.deployments[0]["base-commit-meta"]["rpmostree.packages"]'
+      register: ros
+
+    - name: Indicate 'atomic-registries' is layered
+      when: "'atomic-registries' in ros.stdout"
+      set_fact:
+        ar_is_layered: true
```
mike-nguyen: This assumes deployment 0 is booted, which is likely the case.  Can we resuse the rpm_ostree_status role that sets the booted deployment, then check the layered packages?

```
import_role:
  name: rpm_ostree_status

when: "'atomic-registries' in ros_booted["base-commit-meta"]["rpmostree.packages"]
set_fact:
...
```

tests/atomic/main.yml
```diff
@@ -471,3 +471,46 @@
...
+    - block:
+        - import_role:
+            name: rpm_ostree_rollback
+
+        - import_role:
+            name: reboot
+
+        - name: Cleanup deployments
+          command: rpm-ostree cleanup -rpmb
+      when:
```
mike-nguyen: Is it possible to move then when statement to the top to having consistent styling? 
```
when: 
  - ar_is_layered is defined
  - ar_is_layered
block:
...
```

tests/atomic/main.yml
miabbott: We have high confidence about the state of the system, which is why I hard coded the first deployment in this case.

But I'll adapt it to check for 'booted' since you never know when things will go sideways.

mike-nguyen: Knowing me, I would add a test before it and make it goes sideways :laughing:

tests/atomic/main.yml
miabbott: Ah OK, that is the style we want.  I did a `git grep` for the blocks and we aren't consistent with that...

mike-nguyen: I wasn't sure if it works, but it should?  We can  clean up the other blocks in separate PR.

miabbott:
Damn, all the CI test failed because `import_role` is not supported in Ansible 2.2

We'll have to hold off on merging this until #373 is integrated.

mike-nguyen:
bot, retest this please

miabbott:
Rebased and force pushed the update

mike-nguyen:
dang, it looks like we delete the base-commit-meta entry when we run the rpm_ostree_status role :unamused: We probably have to revert back to what you had originally.

```
TASK [rpm_ostree_status : Get rpm-ostree status output] ************************
changed: [testnode] => {
    "changed": true, 
    "cmd": "rpm-ostree status --json | docker run -i docker.io/miabbott/aht-tools jq 'del(.deployments[][\"layered-commit-meta\"][\"rpmostree.rpmdb.pkglist\"],\n.deployments[][\"base-commit-meta\"][\"rpmostree.rpmdb.pkglist\"])'", 
    "delta": "0:00:01.857563", 
    "end": "2018-04-27 18:45:22.674228", 
    "failed": false, 
    "rc": 0, 
    "start": "2018-04-27 18:45:20.816665"
}
```

miabbott:
> dang, it looks like we delete the base-commit-meta entry when we run the rpm_ostree_status role unamused We probably have to revert back to what you had originally or remove the base-commit-meta from being deleted with jq.

I thought  the delete was only touching the `rpmostree.rpmdb.pkglist`, not the entire `base-commit-meta` dictionary.  (Also, I swear this worked before I pushed it!)

I'll fiddle with it some more; if I have to revert back to the original implementation, NBD.

mike-nguyen:
@miabbott, Let me double check!  Now that i look at it, it should be only deleting rpmostree.rpmdb.pkglist but I didn't see it in ros_booted.  I'll have to dig further up to see whats happening in the test output.

miabbott:
Yeah, I was correct about the individual keys.  The problem is that the `base-commit-meta` dict doesn't have the `rpmostree.packages` key.  I guess the system I originally tested on had that key set for some reason.  It looks like the safer option is to reference the `layred-commit-meta['rpmostree.packages']` key.

```
ok: [cahc] => {                         
    "ros_booted": {            
        "base-checksum": "6bb2ea169d28feeac3506599048b4123f8de63a22184549a8fec6abf6aa9a185", 
        "base-commit-meta": {
            "rpmostree.inputhash": "7d0d31c21ac223ffa6eb26150a96bdad254288a6dfd82a9b11470f505facf915", 
            "version": "7.2017.1147"
        },                  
        "base-local-replacements": [], 
        "base-removals": [], 
        "base-timestamp": 1524855065, 
        "base-version": "7.2017.1147", 
        "booted": true,                                                                                                                                                                                            
        "checksum": "81d17c9878cd514fe4772d0e4c93cf611622a1cf4d33d81f86e5875e91c6612b", 
        "gpg-enabled": false, 
        "id": "centos-atomic-host-81d17c9878cd514fe4772d0e4c93cf611622a1cf4d33d81f86e5875e91c6612b.0", 
        "layered-commit-meta": {
            "rpmostree.clientlayer": true, 
            "rpmostree.clientlayer_version": 3, 
            "rpmostree.packages": [                                                                                                                                                                                                "atomic-registries"                                                                                                                                                                                            ], 
            "rpmostree.removed-base-packages": [],                                                                                                                                                                
            "rpmostree.replaced-base-packages": [], 
            "rpmostree.rpmmd-repos": [                  
...
```

miabbott:
This fixup is the winner, I swear  :arrow_up:

mike-nguyen:
Nice.  LGTM!

PR #388, opened 2018-04-30
reboot/main: increase default timeout

miabbott:
In RHBZ#1572944 it is speculated that /dev/random is taking a longer
amount of time to initialize with entropy in newer Linux kernels.
The observable result of this is that boot times are increased, to the
tune of 3-4m.  Perhaps I'm being overzealous with the 300s timeout,
but we don't have much to lose at this point.

h/t to @jlebon for the probable root cause in projectatomic/papr#97

### DIFF
--- a/roles/reboot/tasks/main.yml
+++ b/roles/reboot/tasks/main.yml
@@ -16,7 +16,7 @@
 # (otherwise ansible will target the localhost)
 - set_fact:
     real_ansible_host: "{{ ansible_host }}"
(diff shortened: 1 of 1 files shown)


jlebon:
LGTM!

mike-nguyen:
LGTM! It should only wait the full 300s if the system doesn't come back.  In most cases the reboot will wait however long the system takes to reboot and have SSH come back up.

PR #389, opened 2018-04-30
Resolve Ansible 2.4  import_role variable issue

mike-nguyen:
There are two issues encountered after the Ansible 2.4 change.  The
first issue is the rpm_ostree_rebase role.  The refspec variable
inside the role seems to get populated with refspec somewhere up
before the rebase call and causes a failure.  This was fixed by using
a prefixed variable name (ror_refpsec) which is a best practice
anyhow.

The second issue is a little more interesting.  After calling the same
role (rpm_ostree_install_verify) multiple times in the rpm-ostree test,
the import_role doesn't seem to be respecting the variable that is
being passed in.  Instead it is using a value from somewhere up before
the same role is called.  This leads to the conclusion that prefix in
the rebase variable name above may not work in all cases.  This was
fixed by using an include_role which is a dynamic include.  We were
trying to avoid using include_role too often because there is a
recursion depth limit if used too many times.

### DIFF
--- a/roles/rpm_ostree_rebase/tasks/main.yml
+++ b/roles/rpm_ostree_rebase/tasks/main.yml
@@ -14,21 +14,21 @@
 - name: Fail if refspec is not defined
   when: refspec is undefined
   fail:
-    msg: "refspec is not defined"
+    msg: "ror_refspec is not defined"
 
 # Default to empty strings for everything but the refspec
 - name: Setup facts and assign defaults if needed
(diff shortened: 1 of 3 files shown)


mike-nguyen:
I believe this is a bug with Ansible.  I will try to figure out a reproducer and file a bug.

miabbott:
This LGTM; I'm just going to restart the F27 test to see if it will pass.

PR #390, opened 2018-05-03
papr:  update for release of F28

miabbott:
Let's start testing the tests against F28 and switch to using the
'updates' stream on Fedora 27 AH.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -1,11 +1,11 @@
 cluster:
   hosts:
     - name: testnode
-      distro: fedora/27/atomic
+      distro: fedora/28/atomic
   container:
     image: registry.fedoraproject.org/fedora:27
 
-context: fedora/27/atomic
+context: fedora/28/atomic
 
 packages:
   - python-virtualenv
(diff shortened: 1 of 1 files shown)


miabbott:
@jlebon could you take a peek when you get a chance?

.papr.yml
```diff
@@ -50,8 +69,6 @@ cluster:
 context: centos/7/atomic
 
 tests:
-  # skipping kernel_avc_denied due to https://bugzilla.redhat.com/show_bug.cgi?id=1536991
-  #  - should be fixed in CentOS 7.5
   # skipping 'default_t' checks until atomic 1.22 lands with (projectatomic/atomic#1185)
-  - ./.test_director --skip-tags kernel_avc_denied,default_file_label
```
jlebon: Minor: should this be a separate commit? I don't see anything about it in the commit msg.

miabbott: Sure, that would be the proper thing to do

miabbott:
Split the commits and force pushed :arrow_up:

jlebon:
LGTM!

PR #391, opened 2018-05-07
admin-unlock: continue to run on failure

mike-nguyen:
Instead of failing early (the default behavior for Ansible), this PR
uses Ansible's block/rescue to catch errors and continue execution.

There are a few major changes in the structure of the test.
- Each tests is broken into individual yaml files as set of tasks
- Each test that will continue on failure is added as an include_task
in the main.yml file in a block/rescue.
- Each test that we want to fail fatally will not be in a block/rescue
- Each test will capture the result and result details in a list of
dictionaries that can be evaluated at the end of the main.yml test
- When an exception is caught in a block/rescue the failed task
details is stored in ansible_failed_task and ansible_failed_result

The goal was to organize the tests better simplify adding new tests to
the test suite.  main.yml is the test suite and individual testa are
in their own yaml file.  main.yml is much uglier and repetitive due to
an Ansible bug [0] where ansible_failed_result and ansible_failed_task
is undefined if used within a role or with includes.  Ideally, there
would be a role that will include_tasks taking a yaml file is a
parameter and do all the test results tracking.  Once the bug [0] is
fixed, a test can be included by just calling the role.

[0] https://github.com/ansible/ansible/issues/29047

### DIFF
--- a/tests/admin-unlock/main.yml
+++ b/tests/admin-unlock/main.yml
@@ -1,5 +1,15 @@
 ---
-# vim: set ft=ansible:
+# vim: set ft=ansible
+#
+# Logging
+#   Running the playbook will automatically create a test result file in the
+#   playbook directory called admin-unlock-results.log.
+#
+#   To disable result logging, run this playbook with the following option:
+#     -e log_results=false
+#
+#   To change the log file name and location run with the following option:
+#     -e result_file=/your/file.log
 #
 # Core Functionality
 #   Verify rpms can installed and uninstalled after ostree admin unlock
@@ -54,7 +64,14 @@
     tests: []
 
   tasks:
+    - name: Set logging
+      set_fact:
+        log_results: true
+        result_file: "{{ playbook_dir }}/admin-unlock-result.log"
+      tags: setup
+
     - include_tasks: 'setup.yml'
+      tags: setup
 
     # TEST
     # Verify installed/removed packages do not persist through reboot with unlock
@@ -65,7 +82,7 @@
       rescue:
         - set_fact:
             tests: "{{ tests + [ { 'name':'Add/Remove Package - Non Persistence', 'result':'Failed', 'result_details': ansible_failed_result } ] }}"
-
+      tags: add_remove_package_non_persistence
 
     # TEST
     # Verify installed/removed packages persist through reboot with hotfix unlock
@@ -76,7 +93,7 @@
       rescue:
         - set_fact:
             tests: "{{ tests + [ { 'name':'Add/Remove Package - Persistence', 'result':'Failed', 'result_details': ansible_failed_result } ] }}"
-
+      tags: add_remove_package_persistence
 
     # TEST
     # Verify rollback to original deployment after hotfix unlock
@@ -87,7 +104,7 @@
       rescue:
         - set_fact:
             tests: "{{ tests + [ { 'name':'Hotfix rollback', 'result':'Failed', 'result_details': ansible_failed_result } ] }}"
-
+      tags: hotfix_rollback
 
     # TEST
     # Verify no packages can be installed wihtout unlock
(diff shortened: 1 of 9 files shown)


mike-nguyen:
I just wanted to open this first implementation of continuing on failure for discussion.  Feedback is greatly appreciated!  I am very open to changing how this works if you have any thoughts or ideas or notice something that I may have missed.  I can put a WIP label on this if needed.

miabbott:
I was able to run this successfully across multiple streams.  Functionally, it is sound and makes sense to me!  The comments I have are mostly about style:

- could the `vim` comment be included at the top of each file as we have in other places?
- is it possible to trim the line length where the `tests` are declared in each block?
- the use of tags isn't consistent.  could the blocks be tagged at a minimum?  ideally, tag as many tasks/roles that makes sense.

Thinking about how this would fit into our existing Jenkins jobs, we may just want to default to always generating the `aht_result_file` since we will likely want to consume that file in order to email the results out.

tests/admin-unlock/main.yml
```diff
@@ -43,338 +43,94 @@
...
+        - set_fact:
+            tests: "{{ tests + [ { 'name':'Unlock twice hotfix error', 'result':'Passed', 'result_details': '' } ] }}"
+      rescue:
+        - set_fact:
+            tests: "{{ tests + [ { 'name':'Unlock twice hotfix error', 'result':'Failed', 'result_details': ansible_failed_result } ] }}"
+
+    # CLEANUP
+    - block:
+        # WRITE RESULTS TO FILE
+        - name: Save result to file
```
miabbott: I think this should use `become: false`; since it is a local action, it shouldn't require privilege to write a file locally.

mike-nguyen:
> could the vim comment be included at the top of each file as we have in other places?

Will do.

> is it possible to trim the line length where the tests are declared in each block?

I tried using `>` and splitting the line up by Ansible didn't like it.  Maybe I escaped it incorrectly?  If you have any ideas I'm all up for making it look nicer.


> the use of tags isn't consistent. could the blocks be tagged at a minimum? ideally, tag as many tasks/roles that makes sense.

I'll try to figure something out with tagging so we can skip tests or run only individual tests. 

> Thinking about how this would fit into our existing Jenkins jobs, we may just want to default to always generating the aht_result_file since we will likely want to consume that file in order to email the results out.

The way we currently generate the file in the jenkins jobs is by defining the AHT_RESULT_FILE environment variable.  The jobs can be modified to run the ansible playbook with -e aht_result_file=./filename to get same result.

I don't have a strong preference on automatically generating the result file.  I chose not to because that is the current default behavior of our tests.  If you would like to automatically generate a file, what do you think of having a default name (for example, admin-unlock-result.log) then have a boolean result_logging set to true?  The user can turn off the logging by just setting result_logging to false.


Longer term, I would like to get rid of the custom code we have in the callback plugin since all that functionality is available to us through Ansible now.

mike-nguyen:
Pushed changes for `vim: set ft=ansible`, basic tagging, and automatically logging to file.

tests/admin-unlock/main.yml
```diff
@@ -54,7 +64,14 @@
     tests: []
 
   tasks:
+    - name: Set logging
```
miabbott: I don't know if it matters, but maybe we should make the setup phase a `pre_tasks`?

tests/admin-unlock/main.yml
```diff
@@ -120,17 +137,24 @@
       rescue:
         - set_fact:
             tests: "{{ tests + [ { 'name':'Unlock twice hotfix error', 'result':'Failed', 'result_details': ansible_failed_result } ] }}"
+      tags:  unlock_twice_hotfix_error
 
     # CLEANUP
     - block:
```
miabbott: In the same vein, make the cleanup phase `post_tasks`?

tests/admin-unlock/main.yml
mike-nguyen: I've always used pre_tasks and post_tasks with roles because you could not mix tasks and roles.  Usually it is just to setup some variables used in the test.  Since everything is a task now, I'm not sure if it is necessary.  The only benefit I can see is that it is more explicit in the order of the tasks.

miabbott: OK, just wanted to have that conversation.

miabbott:
Let's merge this and see how it goes

PR #392, opened 2018-05-09
admin-unlock: always save log

mike-nguyen:
The way it was previously written, the log file would only be written
if there was a failure during the cleanup.  We want to write the file
when the cleanup fails or passes.  Basically use the always clause.


miabbott:
LGTM!

mike-nguyen:
merging

PR #393, opened 2018-05-09
atomic: continue to run on failure

mike-nguyen:
This is a follow-on to #391 for the atomic test suite.  The changes
allow the tests to continue to run on failure using Ansible's
block/rescue/always.

### DIFF
--- a/tests/atomic/images_update.yml
+++ b/tests/atomic/images_update.yml
@@ -4,8 +4,8 @@
 
 - name: Set image name
   set_fact:
-    fq_name: "docker.io/alpine"
-    local_name: "{{ ansible_docker0.ipv4.address}}:5000/alpine"
+    iu_fq_name: "docker.io/alpine"
+    iu_local_name: "{{ ansible_docker0.ipv4.address}}:5000/alpine"
 
 # Testing images update is a bit tricky since in the docker world, tags don't
 # follow any versioning--latest doesn't really mean latest.  Docker and skopeo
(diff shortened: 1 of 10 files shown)


tests/atomic/pull_digest.yml
```diff
@@ -0,0 +1,34 @@
+---
+# set ft=ansible
+#
+
+- name: Set image name
+  set_fact:
+    fq_name: "docker.io/busybox"
+    digest:
```
miabbott: These two facts should probably be prefixed to make them unique to this playbook.

tests/atomic/pull_storage.yml
```diff
@@ -0,0 +1,49 @@
+---
+# set ft=ansible
+#
+
+- name: Set image name
+  set_fact:
+    fq_name: "docker.io/busybox"
```
miabbott: This should be prefixed to make it unique to the playbook

tests/atomic/images_update.yml
```diff
@@ -0,0 +1,132 @@
+---
+# set ft=ansible
+#
+
+- name: Set image name
+  set_fact:
+    fq_name: "docker.io/alpine"
+    local_name: "{{ ansible_docker0.ipv4.address}}:5000/alpine"
```
miabbott: These should be prefixed to make them unique to the playbook

tests/atomic/pull_tags.yml
```diff
@@ -0,0 +1,57 @@
+---
+# set ft=ansible
+#
+
+- name: Set container name
+  set_fact:
+    fq_name: "docker.io/busybox"
+    short_name: "busybox"
+    tag: 1
```
miabbott: These should be prefixed to make them unique to the playbook

miabbott:
I made a couple of comments about prefixing some variables to be unique, but I don't think it is absolutely necessary.

Since the tests are green and everything else LGTM, we could merge this without those changes.

Let me know if you want to change those variables

mike-nguyen:
I'll make these changes and push a fixup.

mike-nguyen:
Pushed fixup with prefixed variables

PR #394, opened 2018-05-11
docker: continue on failure

mike-nguyen:
This doesn't really continue on failure because there is only one main
test but the logic is there to add more tests in.  This is a
continuation of #391 for the docker test suite.

### DIFF
--- a/tests/docker/main.yml
+++ b/tests/docker/main.yml
@@ -1,4 +1,4 @@
-- name: Ostree Admin Unlock - Test Suite
+- name: Docker - Test Suite
   hosts: all
   become: true
 
@@ -19,7 +19,7 @@
       tags: setup
 
     # TEST
-    # Verify installed/removed packages do not persist through reboot with unlock
+    # Basic functional test of docker
     - block:
         - include_tasks: 'functional.yml'
         - set_fact:
--- a/tests/docker/main.yml
+++ b/tests/docker/main.yml
@@ -1,156 +1,50 @@
----
-# vim: set ft=ansible:
-#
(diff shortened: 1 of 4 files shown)


tests/docker/main.yml
```diff
@@ -1,156 +1,50 @@
...
-      tags:
-        - docker_rm__container
-
-    - role: docker_rmi
-      drmi_image_name: "{{ g_httpd_name }}"
-      tags:
-        - docker_rmi
-
-- name: Docker - Cleanup
+- name: Ostree Admin Unlock - Test Suite
```
miabbott: Copy/pasta error

tests/docker/main.yml
```diff
@@ -1,156 +1,50 @@
...
+      set_fact:
+        log_results: true
+        result_file: "{{ playbook_dir }}/docker-result.log"
+      tags: setup
+
+    - include_tasks: 'setup.yml'
+      tags: setup
+
+    # TEST
+    # Verify installed/removed packages do not persist through reboot with unlock
```
miabbott: more copy/pasta errors

tests/docker/setup.yml
```diff
@@ -0,0 +1,28 @@
+---
+# set ft=ansible
+#
+
+- name: Check for docker-latest package if running against docker-latest
```
miabbott: These two tasks can probably be shortened to a single task when `rpm -q docker-latest` is run and it fails if the RC is 1

miabbott:
Mentioned some nits, but otherwise LGTM.  Tests are green so we can merge after any changes you want to make.

PR #395, opened 2018-05-14
ist: only run OpenSCAP test on RHEL

jlebon:
Otherwise we don't have access to the registry anyway.

### DIFF
--- a/tests/improved-sanity-test/main.yml
+++ b/tests/improved-sanity-test/main.yml
@@ -196,7 +196,8 @@
 
     # TEST
     # Validate atomic scan command
-    - role: atomic_scan_verify
+    - when: ansible_distribution == 'RedHat'
+      role: atomic_scan_verify
       tags:
         - atomic_scan_verify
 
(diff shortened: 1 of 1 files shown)


jlebon:
```
13:52:46 < jlebon> miabbott: but would we even have access to pull the container anyway?
13:52:53 < jlebon> strawman: https://github.com/projectatomic/atomic-host-tests/pull/395
13:52:58 < miabbott> jlebon: yeah, there is no paywall
13:53:34 < jlebon> oh wow
13:55:35 < jlebon> miabbott: i can pull it here indeed. i didn't realize this was still status quo
13:57:18 < jlebon> hmm, ok, so maybe the right fix is to skip this test for now
13:57:27 < jlebon> until centos atomic catches up?
13:57:40 < miabbott> agreed
```

miabbott:
Access to the registry isn't a problem, but something is breaking CAHC right now, so we can use this for now.

miabbott:
The CAHC root cause for the failure is CentOS/sig-atomic-buildscripts#329.

And since the `atomic scan` feature is really only used by RHEL folks, I think it is safe to put this restriction on the sanity test.

PR #396, opened 2018-05-15
rsv: skip RPM signature verification on RHCOS

miabbott:
Until we get the kinks worked out


miabbott:
YOLO merging this because...progress

PR #399, opened 2018-05-23
I-S-T: Add support for running against RHCOS

mike-nguyen:
Lets add tagging to the improved-sanity-test to run against RHCOS.
There is an existing cloud_image tag to run against cloud images so
using tag `rhcos` to run against RHCOS seemed like a good idea.  The
roles without tags did not work with RHCOS so it will only test a
subset of the original improved-sanity-test.

Since RHCOS shares the same ansible_distribution as RHEL AH, I had to
use the ansible_distribution_major version to distinguish between
RHCOS and RHELAH in docker_storage_verify and selinux_verify.

Also snuck in a small fix for the rpm_ostree_rebase role.

### DIFF
--- a/roles/docker_storage_verify/vars/redhat-3.yml
+++ b/roles/docker_storage_verify/vars/redhat-3.yml
@@ -1,0 +1,2 @@
+---
+driver: 'overlay2'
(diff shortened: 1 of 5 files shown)


miabbott:
LGTM!  Good catch on those minor fixes

PR #400, opened 2018-05-23
system-containers: use continue-on-failure model

miabbott:
This updates the `system-containers` suite to use the
'continue-on-failure` model that has been used for other test suites.


mike-nguyen:
LGTM!  I like the casing choice on the test name.  I might have to go back and change the other tests in the future.

PR #401, opened 2018-05-23
roles/rpm_ostree_status: remove tools container on exit

miabbott:
When reusing a test system to validate some other changes, I noticed
that there were a pile of `aht-tools` containers on the host.  This
looks like an oversight when I first started using the tools container
to filter out JSON keys; should make things a little more tidy.


mike-nguyen:
LGTM!

PR #403, opened 2018-05-29
docker-build-httpd: continue on failure

mike-nguyen:
Convert docker-build-httpd to use the continue on failure model
proposed by Issue #375.


miabbott:
@mike-nguyen Do you want to bump out the failure check outside of that last block?

mike-nguyen:
Yes, will do and push another fixup!

mike-nguyen:
:arrow_up: Pushed fixup

PR #405, opened 2018-05-30
docker-swarm: explicitly fail based on test results

mike-nguyen:
I don't think Ansible considers handled exceptions as failures even
though it will display failures in the playbook summary.  Lets
explicitly fail if any of the test results have failures.


miabbott:
No more CI lies!  Thanks!

PR #406, opened 2018-05-30
k8-cluster: explicitly fail based on test results

mike-nguyen:
I don't think Ansible considers handled exceptions as failures even
though it will display failures in the playbook summary. Lets
explicitly fail if any of the test results have failures.

### DIFF
--- a/tests/k8-cluster/main.yml
+++ b/tests/k8-cluster/main.yml
@@ -45,3 +45,13 @@
           local_action: copy content={{ tests | to_nice_yaml(indent=2) }} dest={{ result_file }}
           become: false
       tags: cleanup
+
+
+    # Handled exceptions show up as failures in Ansible but the playbook
+    # itself does not return 0, so explicitly fail the test by checking
+    # the test results
+    - name: Explicitly fail based on test results
+      when: item['result']|lower == "failed"
+      fail:
+        msg: "Failure found in test"
+      with_items: "{{ tests }}"


mike-nguyen:
Still lying.
centos/7/atomic is a false positive
fedora/27/atomic is an actual failure

Why do you work differently Ansible?

mike-nguyen:
I moved the check for failures outside the always block and it seems to catch failures correctly now.

miabbott:
Unfortunately, it means that the `k8-cluster` tests is effectively broken  :laughing:

mike-nguyen:
I just ran locally on cahc and it worked? :man_shrugging:

mike-nguyen:
bot, retest this please

mike-nguyen:
I was testing with a newer version of cahc than was on openstack with centos/7/atomic/continuous and it passed.  /etc/containers/registries.conf is missing which is why the tests are failing starting registries service.  

Here is what I tracked down:

`registries.conf` removed from `atomic-registries` April 2, 2018:
https://src.fedoraproject.org/cgit/rpms/atomic.git/tree/atomic.spec#n260

`registries.conf` added to `skopeo-containers` April 2, 2018:
https://src.fedoraproject.org/rpms/skopeo/blob/master/f/skopeo.spec#_385

`skopeo-containers` renamed to `containers-common` May 30, 2018:
https://src.fedoraproject.org/rpms/skopeo/blob/master/f/skopeo.spec#_329

miabbott:
> Since the docker_private_registry works around CentOS/sig-atomic-buildscripts#316 by installing atomic-registries and atomic-registries no longer creates /etc/containers/registries.conf it is causing the centos tests to fail.

So maybe we should be testing for `/etc/containers/registries.conf` instead and installing `containers-common` now?

miabbott:
> So maybe we should be testing for /etc/containers/registries.conf instead and installing containers-common now?

Started exploring this and ran into the following situation - https://github.com/projectatomic/rpm-ostree/issues/1384

If we can get the CAHC cloud image generation fixed (see CentOS/sig-atomic-buildscripts/#333), the workaround shouldn't be necessary anymore

mike-nguyen:
bot, retest this please

mike-nguyen:
[bz#1585735](https://bugzilla.redhat.com/show_bug.cgi?id=1585735) is the exact same error I'm seeing on Fedora 27 atomic (latest) and Fedora 27 updates.

mike-nguyen:
bot, retest this please

mike-nguyen:
After installing https://bodhi.fedoraproject.org/updates/FEDORA-2018-b6ba25b167 referenced in https://bugzilla.redhat.com/show_bug.cgi?id=1585735, the fedora/27/atomic tests passes with this PR.

miabbott:
@mike-nguyen that docker update was obsoleted by https://bodhi.fedoraproject.org/updates/FEDORA-2018-993659ebfd

Could you retry your tests with the new build and give karma to the update if it passes?

mike-nguyen:
>@mike-nguyen that docker update was obsoleted by https://bodhi.fedoraproject.org/updates/FEDORA-2018-993659ebfd
Could you retry your tests with the new build and give karma to the update if it passes?

Added karma!  

I'll look into the centos 7 failure now.

mike-nguyen:
bot, retest this please

mike-nguyen:
bot, retest this please

mike-nguyen:
I think I got mixed up.  CAHC doesn't seem to have atomic-registries so I had to put the workaround back in.

mike-nguyen:
I think this is ready to go.  fedora/27/atomic failure is a known issue that is fixed upstream.

PR #408, opened 2018-06-11
system-containers: explicitly fail on results

mike-nguyen:
Ansible doesn't seem to fail on handled exceptions so lets explicitly
fail based on the test results.


miabbott:
LGTM

PR #410, opened 2018-06-11
admin-unlock: explicitly fail on test results

mike-nguyen:
Ansible does not consider handled exceptions failures and will return
0.  Lets explicitly fail the playbook based on test results so Ansible
will return non-zero.


miabbott:
LGTM

PR #411, opened 2018-06-11
atomic: explicitly fail on test results

mike-nguyen:
Ansible does not consider handled exceptions failures and will return
0.  Lets explicitly fail the playbook based on test results so Ansible
will return non-zero.


mike-nguyen:
These failures look like the same cause as #406

miabbott:
Meh...I'll allow it.

PR #412, opened 2018-06-11
package-layering: explicitly fail on test results

mike-nguyen:
Ansible does not considered handled exceptions failures.  Lets
explicitly fail based on test results.


miabbott:
Thanks for cleaning up all these cases!

PR #413, opened 2018-06-11
treewide: add in correct vim modeline and YAML doc header

miabbott:
The OCD took over and I updated all of the YAML files to start with
the `---` document separator[0] and have a proper `vim` modeline[1]

[0] http://yaml.org/spec/1.0/#id2561718
[1] http://vimdoc.sourceforge.net/htmldoc/options.html#modeline

### DIFF
--- a/roles/ansible_version_check/meta/main.yml
+++ b/roles/ansible_version_check/meta/main.yml
@@ -1,2 +1,3 @@
 ---
+# vim: set ft=ansible:
 allow_duplicates: true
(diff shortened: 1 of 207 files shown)


miabbott:
`centos/7/atomic` and `centos/7/atomic/continuous` failed due to the `registries.conf` mess discussed in #406 

`fedora/27/atomic` was an infra flake

mike-nguyen:
LGTM!  Its all comments so I'm not concerned about the CI pass/fail status.  If one passes, we're golden!

PR #415, opened 2018-06-13
docker_pull_run_remove: add test for cpu-shares flag

### LINKED ISSUE #414: add docker `cpu-shares` test to a-h-t
There was recently a regression in `docker` which impacted the use of `oc cluster up`:

https://bugzilla.redhat.com/show_bug.cgi?id=1585735

The simple test case is to use the `--cpu-shares` flag when doing `docker run`

`docker run --cpu-shares 2 --rm -ti fedora:28 date`

We should add this into the tests; probably the `improved-sanity-test`

miabbott:
This changes the `docker_pull_run_remove` role to add a test for the
use of `--cpu-shares` flag when doing `docker run`.  This is to catch
regressions such as RHBZ#1585735.

Additionally, I expanded the list of images we are testing to include
images from registries other than docker.io.

Closes #414

### DIFF
--- a/roles/docker_pull_run_remove/tasks/main.yml
+++ b/roles/docker_pull_run_remove/tasks/main.yml
@@ -4,7 +4,7 @@
 # popular_images is defined in roles/docker_pull_run_remove/vars/main.yml
 #
 - name: Pull the popular images from Docker Hub
-  command: "docker pull docker.io/{{ item }}"
+  command: "docker pull {{ item }}"
   with_items: "{{ popular_images }}"
   register: dpd
   retries: 5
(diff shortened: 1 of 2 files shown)


miabbott:
The `centos/7/atomic/continuous` check will probably fail due to CentOS/sig-atomic-buildscripts#329

miabbott:
Heh...the Fedora checks might break too, because the new `docker` build hasn't landed in the streams yet.

Let's wait on merging this until this update gets included in the stream:

https://bodhi.fedoraproject.org/updates/FEDORA-2018-993659ebfd

mike-nguyen:
Sounds good.

mike-nguyen:
Sounds good!

PR #416, opened 2018-06-20
dprr: Clean up language and hotfix the centos image

miabbott:
The `name` of the first task in this role wasn't accurate once we
started pulling more images from other registries.  This makes the
language a bit more generic.

Additionally, the r.c.o/centos image recently started to fail and this
was causing all the sanity tests to fail.  Switch the image to use
r.c.o/centos/centos as a hotfix.


mike-nguyen:
LGTM

dustymabe:
+1

PR #417, opened 2018-06-20
introduce podman sanity checks

### LINKED ISSUE #378: add test cases for 'podman'
We are including `podman` in Fedora 28/Rawhide, so we should look at adding some test cases to the sanity test around `podman`.

Additionally, we should investigate if we can run the `docker` test suite we have using `podman`

miabbott:
`podman` is included in most (all?) of the AH streams that we are testing, so let's start making sure it works on Atomic Host.

This is a re-implementation of the `docker_pull_run_remove` role that uses `podman`.  It's slightly more comprehensive than the original role, but still very bare bones.

Closes #378

### DIFF
--- a/roles/podman_pull_run_remove/meta/main.yml
+++ b/roles/podman_pull_run_remove/meta/main.yml
@@ -1,0 +1,3 @@
+---
+# vim: set ft=ansible:
+allow_duplicates: true
--- a/roles/podman_pull_run_remove/tasks/main.yml
+++ b/roles/podman_pull_run_remove/tasks/main.yml
@@ -1,0 +1,56 @@
+---
+# vim: set ft=ansible:
+#
+# This is a copy of the `docker_pull_run_remove` role which has been adapted
+# to use `podman`.  There are some minor changes that expand on the original
+# role to make this more comprehensive.
+#
+# `popular_images` is defined in roles/podman_pull_run_remove/vars/main.yml
+# It is a dict using image names as a key and the value is a command that
+# can be run.
+#
+
+# Check to see if the host has podman, but don't fail if it is not installed
+- name: Check if podman is installed
+  command: rpm -q podman
+  register: podman
+  ignore_errors: true
+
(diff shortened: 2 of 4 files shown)


miabbott:
There is an existing BZ or two about network access when using `podman run` that is going to bite us with this test.

I could add a separate tag that we could filter on or just wait until the BZs are fixed to merge this.

mike-nguyen:
I'm okay with waiting for the BZs.  Can you copy the BZs here so when they're fixed, I can just re-run and merge this?

miabbott:
@mike-nguyen my understanding is that we need this https://github.com/containernetworking/plugins/pull/75 merged and then we need a new build of `podman`

mike-nguyen:
Is this temporarily fixed with this: https://github.com/containers/libpod/pull/1431 ?

miabbott:
It *should* be; but we still have to wait for builds to land across all the platforms.

miabbott:
bot, retest this please

miabbott:
bot, retest this please

miabbott:
CentOS is still going to fail because the build of `podman` is too old to have the fix for `--net=host`, but everything else should be OK.

mike-nguyen:
@miabbott  I'm not too concerned about CentOS7.  CAHC seems to be failing pulling from registry.access.redhat.com because it doesn't have a cert.  If one is required can we put one in or just skip the r.a.rh.com test completely?

miabbott:
Yeah, the root cause is here:

https://github.com/CentOS/sig-atomic-buildscripts/issues/329

I think there is still value in pulling from `registry.access.redhat.com`, so I'll further adjust the test to skip that registry on CAHC

miabbott:
Just pushed a new commit with fixes.  :arrow_up: 

The CI checks are probably going to fail because OpenStack is bad and should feel bad.

mike-nguyen:
Look, it worked!

mike-nguyen:
LGTM!

PR #418, opened 2018-06-26
dprr: modify list of "popular" images for CentOS based streams

miabbott:
The i-s-t has been failing for longer than I would care to admit
because of CentOS/sig-atomic-buildscripts#329 and there's been little
movement in the direction of fixing that.  So we'll workaround the
deficincieny by using two lists of images, one that works for CentOS
derived streams and another that works for the rest.


miabbott:
`centos/7/atomic` - failed because `firewalld` is now included (need to update that test)

`centos/7/atomic/continuous` - failed due to unrelated issue

mike-nguyen:
LGTM.  I'm going to merge this  because CI failures aren't related to this.

PR #420, opened 2018-06-28
README.md: update dashboard link to FAHC results

jlebon:


mike-nguyen:
LGTM.  @miabbott  are you okay with merging before an actual fahc28 run?  It will just show up in the readme as Fedora 28 Continuous but with a badge that says 27 in the version.

PR #419, opened 2018-06-26
fsv: always check for the presence/usage of firewalld

miabbott:
All of the Atomic Host streams should now have `firewalld` installed,
so let's remove the conditionals and always check that it is there.

PR #421, opened 2018-06-28
dsv: update role to be more consistent across all streams

miabbott:
Almost all of the streams are using `overlay2` as the docker storage
driver now, so we can remove a lot of the conditional workarounds we
had.

If we can get CAHC to switch to `overlay2` (see
CentOS/sig-atomic-buildscripts#341), we can simplify this role even
more.

### DIFF
--- a/roles/docker_storage_verify/tasks/main.yml
+++ b/roles/docker_storage_verify/tasks/main.yml
@@ -1,33 +1,19 @@
 ---
 # vim: set ft=ansible:
 #
-- name: Include distribution vars
-  include_vars: "{{ item }}"
-  with_first_found:
-    - "{{ ansible_distribution|lower }}-{{ ansible_distribution_version }}.yml"
-    - "{{ ansible_distribution|lower }}-{{ ansible_distribution_major_version }}.yml"
-    - "{{ ansible_distribution|lower }}.yml"
+- name: Set storage driver
+  set_fact:
+    storage_driver: 'overlay2'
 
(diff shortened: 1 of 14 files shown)


miabbott:
LOL...the `centos` checks appear to be failing due to the `firewalld_service_verify` role, which is fixed in #419 

`fedora/27/atomic` is failing due to a [known docker issue](https://bugzilla.redhat.com/show_bug.cgi?id=1595300)

Some chicken and egg stuff here

roles/docker_storage_verify/tasks/main.yml
```diff
@@ -1,33 +1,19 @@
...
+- name: Set storage driver
+  set_fact:
+    storage_driver: 'overlay2'
 
-- name: Fail if required variables are not defined
-  when: driver is undefined
-  fail:
-    msg: "The required variables are not defined.  Check files under roles/docker_storage_verify/vars"
+- when: ansible_distribution == 'CentOSDev'
+  name: Set storage driver for CAHC
```
mike-nguyen: I know we are trying to put the when statement first stylistically, but i think it makes more sense for the name to go first if there is a name, then the actual code underneath.

mike-nguyen:
These are both fairly low risk.  I am okay with merging them back to back!

roles/docker_storage_verify/tasks/main.yml
miabbott: no big deal, i can make that change

PR #423, opened 2018-07-11
i-s-t: remove rhcos tag from firewalld check

mike-nguyen:
Lets remove the rhcos tag from the firewalld role because rhcos does
not have firewalld.


miabbott:
I'm OK with this, but I think we should also ask if `firewalld` **should** be part of RHCOS

PR #424, opened 2018-07-11
new 'oc cluster up' test case

### LINKED ISSUE #407: Sanity check `oc cluster up` on streams that have OpenShift built-in
This would be a nice check to have for RHCOS.

miabbott:
This adds a basic test to see if `oc cluster up` works on Atomic
Hosts.

All the AH streams, except for RHEL, package layer the
`origin-clients` package and then do some real simple `oc` commands.
For RHEL, it was easier to just copy the `oc` binary into
`/usr/local/bin` than try to figure out the right entitlements to
install the equivalent `origin-clients` package.

Closes #407


tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+        name: rpm_ostree_install
+      vars:
+        roi_packages: "origin-clients"
+        roi_reboot: false
+
+    - import_role:
+        name: reboot
+
+# I couldn't find a straight-forward way to get a copy of `origin-clients`
+# for RHEL, so I'm just going to use the binary from GitHub
```
jlebon: For RHEL, I think it's called `atomic-openshift-clients`.

tests/oc-cluster-up/cleanup.yml
```diff
@@ -0,0 +1,49 @@
...
+    state: absent
+
+# Remove insecure registry
+- name: Remove the insecure registry
+  replace:
+    regexp: "^registries = \\['172.30.0.0/16'\\]"
+    replace: "registries = []"
+    dest: /etc/containers/registries.conf
+
+- name: Restart docker daemon
```
miabbott: Hmm, I should restart `registries` here too

tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+    state: restarted
+  with_items:
+    - registries
+    - docker
+
+# Drop in the openshift-release repo + key for CentOS
+- when: "'CentOS' in ansible_distribution"
+  block:
+    - name: Grab the OpenShift Origin repo
+      command: curl -LO https://raw.githubusercontent.com/CentOS-PaaS-SIG/centos-release-openshift-origin/master/CentOS-OpenShift-Origin.repo
```
mike-nguyen: get_url module instead of curl?

tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+      args:
+        chdir: /etc/yum.repos.d/
+
+    - name: Grab the CentOS SIG PaaS GPG key
+      command: curl -LO https://raw.githubusercontent.com/CentOS-PaaS-SIG/centos-release-paas-common/master/RPM-GPG-KEY-CentOS-SIG-PaaS
```
mike-nguyen: same here

tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+# I couldn't find a straight-forward way to get a copy of `origin-clients`
+# for RHEL, so I'm just going to use the binary from GitHub
+- when: ansible_distribution == "RedHat"
+  block:
+    - name: Create temp directory
+      command: mktemp -d
+      register: oso_temp
+
+    - name: Grab OpenShift Origin binary
+      command: curl -LO https://github.com/openshift/origin/releases/download/v3.9.0/openshift-origin-client-tools-v3.9.0-191fece-linux-64bit.tar.gz
```
mike-nguyen: same here and maybe rename it something shorter so it is easier to reference below when you unarchive it?

tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+      args:
+        chdir: "{{ oso_temp.stdout }}"
+
+    - name: Extract binary
+      command: tar zxvf openshift-origin-client-tools-v3.9.0-191fece-linux-64bit.tar.gz
```
mike-nguyen: unarchive module instead of command?

tests/oc-cluster-up/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+      args:
+        chdir: "{{ oso_temp.stdout }}"
+
+    - name: Copy oc binary to /usr/local/bin
```
mike-nguyen: copy module? i think you have to set `remote_source=true` if you use this

tests/oc-cluster-up/setup.yml
miabbott: @mike-nguyen I stuck with the bare file name because when the renamed archive is unpacked, you still get the long-ass name.  It wasn't worth it to me to try to fight it.

PR #426, opened 2018-07-24
initramfs: update test for BootLoaderSpec change

mike-nguyen:
The BootLoaderSpec filename change in ostree [0] is causing failures
in streams containing the change.  This commit will handle pre and
post BLS filename changes using a simple regex.

[0] https://github.com/ostreedev/ostree/pull/1654

### DIFF
--- a/tests/rpm-ostree/initramfs.yml
+++ b/tests/rpm-ostree/initramfs.yml
@@ -54,7 +54,7 @@
     osname: "{{ ros_booted['osname'] }}"
 
 - name: Get boot entry file
-  shell: ls -1 /boot/loader/entries/ | grep -E "ostree-[2-]*fedora-atomic[-0]*.conf"
+  shell: ls -1 /boot/loader/entries/ | grep -E "ostree-[2-]*{{ osname }}[-0]*.conf"
   register: boot_file
 
 - name: Get initrd file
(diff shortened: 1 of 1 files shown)


mike-nguyen:
centos/cahc are failing for unrelated reasons

tests/rpm-ostree/initramfs.yml
```diff
@@ -53,9 +53,13 @@
   set_fact:
     osname: "{{ ros_booted['osname'] }}"
 
-- name: Get bootloader entry
+- name: Get boot entry file
+  shell: ls -1 /boot/loader/entries/ | grep -E "ostree-[2-]*fedora-atomic[-0]*.conf"
```
miabbott: Is this going to work for non Fedora streams?

mike-nguyen: :man_facepalming: I was testing with fedora

cgwalters:
Related, looks like this test overlaps a lot with the [upstream one](https://github.com/projectatomic/rpm-ostree/blob/master/tests/vmcheck/test-initramfs.sh) - which was [adapted for the BLS change](https://github.com/projectatomic/rpm-ostree/commit/bc52e8eaf36976b9c78b1fdd5131cab2f352d059#diff-efcd54ddb08caf367d1569ff99ad6f3a).

Which leads back to [ye olde discussion](https://github.com/projectatomic/atomic-host-tests/issues/74).

miabbott:
Tested locally across all the supported streams without issue!

PR #427, opened 2018-07-24
i-s-t: set container_manage_cgroup for all streams

mike-nguyen:
CentOS Atomic Host and CAHC inherited the systemd container cgroups
issue and now needs the container_manage_cgroup seboolean.


miabbott:
Makes sense to me.  Merging!

PR #429, opened 2018-07-25
auv: use container name in uninstall

mike-nguyen:
CentOS AH has an old version of atomic which is fixed upstream [0]
which causes the uninstall of cockpit to fail if a name is not passed
in.  This commit works around it until CentOS Atomic Host is updated.

[0] https://github.com/projectatomic/atomic/issues/1219


miabbott:
LGTM!

PR #431, opened 2018-08-27
stop using workarounds for `atomic` command

miabbott:
The `atomic` command is fixed across all the streams, so we can
stop using the workarounds.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -53,8 +53,7 @@ cluster:
 context: centos/7/atomic/continuous
 
 tests:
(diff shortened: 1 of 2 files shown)


mike-nguyen:
centos/7/atomic still too old?

miabbott:
The version in OpenStack is, but the latest released version should be fine.

Once we get the image updated, I'll re-run the CI for that.

miabbott:
@mike-nguyen all green!  :smile:

miabbott:
I left that in because we get coverage on the use of `-n` during `atomic install` and with it omitted later on.

mike-nguyen:
LGTM then

PR #432, opened 2018-08-30
selinux_verify: update vars file name for rhcos

mike-nguyen:
The redhat-release-coreos package bumped the os-release version to 4
the file needs to be renamed so the selinux_verify role will pick up
the correct vars.


miabbott:
LGTM

PR #433, opened 2018-09-24
aiv: use name when uninstalling container

miabbott:
On RHELAH, it seems that you need to pass the name of the container
when using `atomic uninstall` if `atomic install` has been used
multiple times.  (This only seems to bite me when I'm reusing a host
for multiple rounds of tests and the host is not "clean".  The
automated jobs appear to avoid this condition.)


cgwalters:
LGTM

PR #437, opened 2018-10-03
handler: also save journal of previous boot

jlebon:
This can be help when an error in a current boot happens because of
something that happened in the *previous* boot. There doesn't seem to be
a concise way to tell `journalctl` to list both the current and previous
boots but not the ones before that, so we just append to the tmpfile in
a second invocation.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -21,6 +21,9 @@ tests:
   # a docker-latest
   - ./.test_director -e g_docker_latest=false
 
+artifacts:
+  - tests/improved-sanity-test/improved-sanity-test-testnode-journal
+
 ---
 inherit: true
 
--- a/roles/handler_notify_on_failure/handlers/main.yml
+++ b/roles/handler_notify_on_failure/handlers/main.yml
@@ -31,6 +31,11 @@
   listen: h_get_journal
   when: aht_result == "1"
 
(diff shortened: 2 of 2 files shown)


miabbott:
LGTM

@mike-nguyen You did the original implementation...could you merge it if you likey?

miabbott:
@jlebon interesting that the CI check for F29 is showing the same failure we are trying to debug...

jlebon:
Yeah, added the journal as an artifact to try to debug it right here!

jlebon:
OK, so now we have logs:

```
Oct 03 16:28:20 testnode systemd[1]: Stopped OSTree Finalize Staged Deployment.
Oct 03 16:28:20 testnode systemd[1]: ostree-finalize-staged.service: Failed with result 'exit-code'.
Oct 03 16:28:20 testnode systemd[1]: ostree-finalize-staged.service: Control process exited, code=exited status=1
Oct 03 16:28:20 testnode ostree[8354]: error: Bootloader write config: grub2-mkconfig: Child process exited with code 1
```

Though no clear indication what the error actually was. So at this point, we'd boot up an f29 and try it out there to dig deeper.

But at least we know the journal gathering works!

mike-nguyen:
LGTM!

PR #444, opened 2018-10-16
README: point to latest results for f28 updates

miabbott:


miabbott:
I muffed the change in #442, so this is the right fix

mike-nguyen:
LGTM!

PR #446, opened 2018-12-12
roles: add regression tests for docker/podman exec

miabbott:
We haven't been testing the `docker/podman exec` path and missed the
following regressions in RHELAH:

https://bugzilla.redhat.com/show_bug.cgi?id=1650512
https://bugzilla.redhat.com/show_bug.cgi?id=1655214

This adds a simple test for both.

### DIFF
--- a/roles/docker_pull_run_remove/tasks/main.yml
+++ b/roles/docker_pull_run_remove/tasks/main.yml
@@ -30,15 +30,15 @@
 
 # Test for https://bugzilla.redhat.com/show_bug.cgi?id=1655214
 - name: Run popular images as detatched
-  command: "docker run --name test-{{ item.split('/')[1] }} -d --rm {{ item }} sleep 1000"
+  command: "docker run --name test-{{ item.split('/')[-1] }} -d --rm {{ item }} sleep 1000"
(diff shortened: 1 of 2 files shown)


roles/docker_pull_run_remove/tasks/main.yml
```diff
@@ -28,6 +28,19 @@
   command: "docker run --cpu-shares 2 --rm {{ item }} echo 'hello'"
   with_items: "{{ dprr_images }}"
 
+# Test for https://bugzilla.redhat.com/show_bug.cgi?id=1655214
+- name: Run popular images as detatched
+  command: "docker run --name test-{{ item.split('/')[1] }} -d --rm {{ item }} sleep 1000"
```
mike-nguyen: Can we use `item.split('/')[-1]` instead of `item.split('/')[1]`?  If for some reason in the future we have a `dprr_images` without `/` in it, index 1 would be out of range where -1 will still be valid.  Same comment for the variations below.

miabbott: Smart.  Will update.

mike-nguyen:
LGTM

PR #448, opened 2019-01-11
livefs: adapt test to new behavior

mike-nguyen:
Livefs has been reworked and the deployments after rebooting from
livefs has changed to only contain the livefs deployment and the
finalized staged deployment.  The original deployment (pre-layering
and livefs) will no longer be in the deployment list.


miabbott:
Ran the `livefs` portion of the `rpm-ostree` tests and looked good here.  Thanks for fixing this!

PR #449, opened 2019-01-14
atomic: handle multiple registries shortname pull

mike-nguyen:
The image cockpit/ws matches images in docker.io and
registry.centos.org.  When pulling with short name, it is not
predictable which registry atomic will pull from.  This change is a
bit of a hack to force atomic to pull from registry.centos.org.


miabbott:
Hacks abound!  LGTM

PR #450, opened 2019-01-16
rpm_ostree_livefs: detect which version of livefs cmd to use

miabbott:
Distros that don't rev their version of `rpm-ostree` as frequently are
missing the `--i-like-danger` switch when using `rpm-ostree ex
livefs`.

This change detects the version of `rpm-ostree` and applies the switch
appropriately.

### DIFF
--- a/roles/rpm_ostree_livefs/tasks/main.yml
+++ b/roles/rpm_ostree_livefs/tasks/main.yml
@@ -1,8 +1,28 @@
 ---
 # vim: set ft=ansible:
 #
+- name: Determine rpm-ostree version
+  block:
+    - command: rpm-ostree --version
+      register: ros_out
+
+    - set_fact:
+        ros_yaml: "{{ ros_out.stdout | from_yaml }}"
+
+    - set_fact:
+        ros_version: "{{ ros_yaml['rpm-ostree']['Version'] }}"
+
+- name: Set old livefs command
+  set_fact:
+    ros_livefs_cmd: "rpm-ostree ex livefs"
+
+- name: Set new livefs command
+  when: ros_version | version_compare('2018.9', '>=')
+  set_fact:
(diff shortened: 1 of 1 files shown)


mike-nguyen:
Code looks good, unfortunately I think it will fail on RHEL AH because it probably doesn't behave the same was as the changes I made in #448.  I haven't tested this theory.

jlebon:
IMO, I'm not sure it's worth spending too much time trying to get livefs tests to pass. It's known broken right now, its use is discouraged, and it's likely to change semantics in the future.

mike-nguyen:
@miabbott  I'm okay with merging this and then just letting it be a known failure in RHEL AH.

miabbott:
> Code looks good, unfortunately I think it will fail on RHEL AH because it probably doesn't behave the same was as the changes I made in #448. 

The whole point of this change was to get it to work on RHELAH.  :thinking:

miabbott:
> The whole point of this change was to get it to work on RHELAH. thinking

I'm pretty sure I tested this before submitting the PR, but it was late last night, so I could have been asleep.

Let me double-check my tests...

mike-nguyen:
Just to re-iterate, I didn't test this theory :laughing:

miabbott:
> Just to re-iterate, I didn't test this theory laughing

It's been a rough morning already...my reading comprehension is suffering  :roll_eyes:

miabbott:
Or we just take the out from @jlebon and remove the `livefs` test completely...

mike-nguyen:
@miabbott I'm fine with removing the test.

miabbott:
YOLO...going to just remove it.

Closed in favor of #451

miabbott:
FWIW, these changes worked on RHELAH  ¯\_(ツ)_/¯

PR #451, opened 2019-01-16
tests/rpm-ostree:  remove livefs test

miabbott:
This removes the `livefs` portion of the test and changes the
`rpm_ostree_livefs` role to fail if you try to use it.


miabbott:
> LGTM. We can always add the test back in since livefs.yml is still there.

Oh, I forgot about that.  Pushed a commit to remove it :arrow_up:

mike-nguyen:
LGTM!

PR #452, opened 2019-02-04
roles/user_verify*: change grep pattern for home dir change

miabbott:
It looks like projectatomic/rpm-ostree#1726 changed how `useradd`
operates on `ostree` hosts, so we have to look for `/var/home/<user>`
now.  Except that not all the `ostree` hosts will have this change, so
make it an optional part of the regex pattern.


miabbott:
Hmm...this worked in my local tests.  I'll have another look.

mike-nguyen:
LGTM

PR #453, opened 2019-04-08
rpm-ostree: update tzdata package

mike-nguyen:
The Fedora 27 tzdata package location no longer exists so use the
Fedora 29 tzdata package instead.


cgwalters:
Looks like it's about time!

miabbott:
> Looks like it's about time!

/me groans

PR #454, opened 2019-04-25
system-containers: update hello-world image

mike-nguyen:
The original hello-world image broke so a new image was built with
the original Dockerfile and uploaded to docker hub.  This file
should be used going forward.


mike-nguyen:
> @mike-nguyen Not entirely necessary, but nice to have...is that image set up to automatically build on a new git commit somewhere?

@miabbott as of now, no.  I can make it happen though.  The original image was setup like that.  For some reason the image was updated but when I looked at the source repo, nothing changed.  I think the owner may have pushed a new container for testing manually.  This one is at least under our control and known to be working.

PR #457, opened 2019-07-10
roles/selinux_verify: remove docker-latest from RHEL

miabbott:
All the docker-latest bits will be removed in 7.6.z, so update
accordingly.

Added a missing file from `/usr/libexec/docker` that was overlooked


mike-nguyen:
LGTM

PR #460, opened 2020-08-03
Updates to container image registries

miabbott:
We saw some rate limiting issues when pulling from docker.io, so these are some changes to help alleviate that problem in the future


mike-nguyen:
LGTM

